PR #10, opened 2021-08-25
Refine symbol declaration hierarchy and store symbols in the corresponding trees

cache-nez:
This PR trims down the symbol tree: now it consists only of declarations, i.e. symbols that can be selected. All symbols are now also stored in the corresponding trees.

### DIFF
--- a/jvm/src/test/scala/BaseUnpicklingSuite.scala
+++ b/jvm/src/test/scala/BaseUnpicklingSuite.scala
@@ -1,0 +1,21 @@
+import tastyquery.Contexts
+import tastyquery.Contexts.Context
+import tastyquery.ast.Trees.Tree
+import tastyquery.ast.Types.Type
+import tastyquery.reader.TastyUnpickler
+
+import java.nio.file.{Files, Paths}
+
+abstract class BaseUnpicklingSuite extends munit.FunSuite {
+  val ResourceProperty = "test-resources"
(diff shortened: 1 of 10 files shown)


jvm/src/test/scala/BaseUnpicklingSuite.scala
```diff
@@ -0,0 +1,21 @@
...
+import java.nio.file.{Files, Paths}
+
+abstract class BaseUnpicklingSuite extends munit.FunSuite {
+  val ResourceProperty = "test-resources"
+
+  def unpickle(filename: String)(using ctx: Context = Contexts.empty): Tree = {
+    val resourcePath = getResourcePath(filename)
+    val bytes        = Files.readAllBytes(Paths.get(resourcePath))
+    val unpickler    = new TastyUnpickler(bytes)
+    unpickler.unpickle(new TastyUnpickler.TreeSectionUnpickler()).get.unpickle (using ctx).head
```
bishabosha: did scalafmt add this space?

cache-nez: Yes, with the dotty runner. Just updated the runner to scala3 and it's all pretty now!

bishabosha: nicee

PR #30, opened 2022-01-31
Addition of Classfile parser

bishabosha:
Adds classpath scanning and loading of classes up to:
- java field and method names
- scala 2 member symbols

no type information is added yet.

Also adapted code to be null safe


bishabosha:
ok done with the small changes

jvm/src/main/scala/Main.scala
```diff
@@ -1,32 +1,36 @@
...
       case "--standalone" :: filename :: Nil =>
-        val unpickler = new TastyUnpickler(Files.readAllBytes(Paths.get(filename)))
-        println(
-          unpickler.unpickle(new TastyUnpickler.TreeSectionUnpickler()).get.unpickle(using Contexts.empty(filename))
-        )
+        import unsafe.unlock.given unsafe.CanObserveIArray[Byte]
+        val unpickler = new TastyUnpickler(unsafe.asArray(IO.readAllBytes(filename)))
+        val ctx = Contexts.empty(Classpath.Empty, filename)
+        println(unpickler.unpickle(new TastyUnpickler.TreeSectionUnpickler()).get.unpickle(using ctx))
+      case "-usejavacp" :: Nil =>
```
sjrd, line 24: Is this really necessary? The availability of this option is already a historical liability in the compiler. Shouldn't we be able to always use `-cp` and explicitly provide the classpath instead?

jvm/src/main/scala/tastyquery/jdk/ClasspathSources.scala
```diff
@@ -0,0 +1,22 @@
...
+import java.io.File
+
+object ClasspathSources {
+
+  def classloaderURLs: IArray[String] = {
+    import scala.language.unsafeNulls
+    val cl = ClassLoader.getSystemClassLoader
+    val urls = cl.asInstanceOf[java.net.URLClassLoader].getURLs
+    val files = urls.map(_.getFile)
+    IArray.unsafeFromArray(files)
```
sjrd, line 12: I doubt the performance requirements of this method warrant the usage of unsafe API.

jvm/src/main/scala/tastyquery/jdk/ClasspathSources.scala
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  def userClasspath: IArray[String] = {
+    import scala.language.unsafeNulls
+    val splitter = File.pathSeparator
+    val paths = sys.props("java.class.path").split(splitter)
+    IArray.unsafeFromArray(paths)
```
sjrd, line 19: Same here.

sjrd (commented):
It seems that there are quite a number of unrelated changes in this PR, scattered around lots of small commits. Would it be possible to decompose it into meaningful commits/PRs? I identified at least

* Null safety of the codebase
* Getting rid of stray files
* Classpath management
* Reading Java .class files
* Reading Scala 2 pickles

It seems that each of those would fit in a dedicated commit, at least. It would help reviewing a lot.

shared/src/main/scala/tastyquery/ast/Symbols.scala
```diff
@@ -37,29 +55,40 @@ object Symbols {
...
-  abstract class DeclaringSymbol(override val name: Name, override val owner: Symbol) extends Symbol(name, owner) {
+  abstract class DeclaringSymbol(override val name: Name, rawowner: Symbol | Null) extends Symbol(name, rawowner) {
     /* A map from the name of a declaration directly inside this symbol to the corresponding symbol
      * The qualifiers on the name are not dropped. For instance, the package names are always fully qualified. */
-    protected val myDeclarations: mutable.HashMap[Name, Symbol] = mutable.HashMap[Name, Symbol]()
 
-    def addDecl(decl: Symbol): Unit = myDeclarations(decl.name) = decl
-    def getDecl(name: Name): Option[Symbol] = myDeclarations.get(name)
-    def declarations: List[Symbol] = myDeclarations.values.toList
+    protected val decls: mutable.MultiDict[Name, Symbol] = mutable.MultiDict.empty
```
sjrd, line 64: This seems to be the only real usage of scala-collection-contrib. Is it really worth the addition of that dependency, as opposed to using a `mutable.Map[Name, mutable.Set[Symbol]]`?

shared/src/main/scala/tastyquery/ast/Symbols.scala
```diff
@@ -37,29 +55,40 @@ object Symbols {
...
+
+    def uninitialised = decls.isEmpty // TODO: hack until decide how to initialise classes lazily
```
sjrd, line 66:
```suggestion
    def uninitialised: Boolean = decls.isEmpty // TODO: hack until decide how to initialise classes lazily
```
*Always* use explicit result types for public and protected methods. This is important for binary compatible evolutions of the library.

jvm/src/main/scala/Main.scala
bishabosha: I haven't tested this main class at all, its just leftovers from prior work, I'd prefer to get rid of it for now

shared/src/main/scala/tastyquery/ast/Symbols.scala
bishabosha: I also use it to sort packages in ClassLoaders

jvm/src/main/scala/tastyquery/jdk/ClasspathSources.scala
bishabosha: I am using IArray as the main collection type throughout, I didn't think this api wasn't recommended, just that it should not be used if you do not know how the underlying array could be shared.

bishabosha:
I can definitely split the work into more structured commits and then repush

jvm/src/main/scala/tastyquery/jdk/ClasspathSources.scala
sjrd: It's better to use `files.toIArray` (if that exists) or `IArray.from(files)` when the path is not performance-critical.

bishabosha:
@sjrd should be ready to review again. I am not sure what is causing the failure with symbolic packages, "works on my machine :)"

shared/src/main/scala/tastyquery/ast/Symbols.scala
bishabosha: removed all usages

bishabosha:
oh no this is terrible, it seems sometimes in TASTy, a package `symbolic_>>` can appear as `symbolic_$greater$greater`, and sometimes as `symbolic_>>`, but I'm not sure how to recreate this. But I also heard that `NameTransformer` isn't idempotent, so I am uncomfortable with always encoding names from classpath :/ @sjrd what do you think?

sjrd:
I would say remove/disable the offending test, and raise an issue in the dotty repo. It is definitely *not* valid TASTy if it contains mangled names in some cases. We shouldn't be dealing with invalid TASTy in this project.

bishabosha:
@sjrd I have removed the assumption of encoded package names, this is good to review

shared/src/main/scala/tastyquery/ast/Names.scala
```diff
@@ -80,15 +86,20 @@ object Names {
...
+    private var myTypeName: TypeName | Null = null
 
     override def toTypeName: TypeName = {
-      if (myTypeName == null) {
-        synchronized {
-          if (myTypeName == null) myTypeName = new TypeName(this)
-        }
+      val local = myTypeName
+      if local == null then synchronized {
+        val local1 = myTypeName
```
sjrd: Consider calling them `local1` and `local2`, since neither is really more important than the other.

Also, consider putting the `synchronized` block on the next line, otherwise it appears you're mixing `if then` with `{}`, which is a bit confusing.

jvm/src/main/scala/Main.scala
```diff
@@ -2,18 +2,23 @@ import java.nio.file.{Files, Paths}
...
         val reader = new ProjectReader
-        reader.read(classpath)
+        val cpElems = ClasspathLoaders.splitClasspath(cp)
+        val classpath = ClasspathLoaders.read(cpElems, Set(ClasspathLoaders.FileKind.Tasty))
+        given BaseContext = Contexts.empty(classpath)
+        reader.read(classes*)
       case _ =>
         println("""Usage:
-                  |-cp <path: String>[ <path: String>[ <path: String>...]]""".stripMargin)
+                  |-cp <paths> <classes>""".stripMargin)
```
sjrd, line 22:
```suggestion
                  |-cp <paths> <classes>...""".stripMargin)
```

jvm/src/test/scala/SymbolSuite.scala
```diff
@@ -48,93 +37,134 @@ class SymbolSuite extends BaseUnpicklingSuite {
...
+      empty_class,
+      Set(empty_class / tname"EmptyClass", empty_class / tname"EmptyClass" / name"<init>")
     )
   }
 
-  test("inner-class") {
-    val ctx = getUnpicklingContext("simple_trees/InnerClass")
-    // Inner is a declaration in InnerClass
-    assertContainsDeclaration(
+  test("basic-symbol-structure-symbollic-names".ignore) {
```
sjrd, line 67:
```suggestion
  test("basic-symbol-structure-symbolic-names".ignore) {
```

sjrd: And the same everywhere else. "symbollic" is a common misspelling.

shared/src/main/scala/tastyquery/ast/Names.scala
```diff
@@ -86,21 +92,27 @@ object Names {
...
-        val local1 = myTypeName
-        if local1 == null then
-          val asType = TypeName(this)
-          myTypeName = asType
-          asType
-        else local1
-      } else {
+      if local == null then
+        synchronized {
+          val local1 = myTypeName
```
sjrd, line 136: Ah well, this change should be in the previous commit.

shared/src/main/scala/tastyquery/ast/Symbols.scala
```diff
@@ -39,7 +39,8 @@ object Symbols {
 
   final class RegularSymbol(override val name: Name, override val owner: Symbol | Null) extends Symbol(name, owner)
 
-  abstract class DeclaringSymbol(override val name: Name, override val owner: Symbol | Null) extends Symbol(name, owner) {
+  abstract class DeclaringSymbol(override val name: Name, override val owner: Symbol | Null)
+      extends Symbol(name, owner) {
```
sjrd: Spurious change here.

shared/src/main/scala/tastyquery/reader/classfiles/Classpaths.scala
```diff
@@ -0,0 +1,54 @@
+package tastyquery.reader.classfiles
+
+import tastyquery.ast.Names.{SimpleName, TermName, EmptyPackageName, termName}
+
+object Classpaths {
+
+  transparent inline def universe(using universe: Universe): Universe = universe
```
sjrd, line 14: Why is this `transparent`? Can't you achieve the same thing with
```suggestion
  def universe(using universe: Universe): universe.type = universe
```
?

shared/src/main/scala/tastyquery/reader/TastyUnpickler.scala
```diff
@@ -12,6 +13,11 @@ import scala.collection.mutable
 
 object TastyUnpickler {
 
+  private def readerFromBytes(bytes: IArray[Byte]): TastyReader = {
+    import unsafe.unlock.given unsafe.CanObserveIArray[Byte]
+    new TastyReader(unsafe.asArray(bytes))
```
sjrd, line 18: This `CanObserveIArray` thing is completely overkill if it's used exactly once, in a private method.

Use
```suggestion
  new TastyReader(bytes.asInstanceOf[Array[Byte]])
```
instead.

In fact, it is not actually safe to this anyway. This is called from the public constructor of `TastyUnpickler`. So we *must* make a copy to be safe.

jvm/src/main/scala/tastyquery/jdk/ClasspathLoaders.scala
```diff
@@ -0,0 +1,194 @@
...
+
+object ClasspathLoaders {
+
+  enum FileKind {
+
+    case Class, Tasty
+
+    val ext: String = {
+      import scala.language.unsafeNulls
+      this.toString.toLowerCase
```
sjrd, line 30: Using `toLowerCase` like that is not safe wrt. systems configured in Turkish, Azeri or Lithuanian. You must pass in `(java.util.locale.ROOT_LOCALE)`. But that doesn't link on Scala.js by default. Prefer
```scala
val ext: String = this match
  case Class => "class"
  case Tasty => "tasty"
```
or something like that.

shared/src/main/scala/tastyquery/reader/classfiles/Classpaths.scala
bishabosha: the choice for transparent inline is that you can use `universe` in an import prefix

shared/src/main/scala/tastyquery/reader/TastyUnpickler.scala
bishabosha: tasty unpickler does not mutate the array, its only unfortunate that it uses array in its api

sjrd: The a cast right there, accompanied by an appropriate comment, should be the right thing to do.

bishabosha: ok - there will be plenty of times where we want to convert between `IArray` an `Array` in the later commits - I thought this import makes it more standardised rather than having casts everywhere

shared/src/main/scala/tastyquery/util/package.scala
```diff
@@ -0,0 +1,21 @@
...
+
+inline def loop(times: Int)(inline op: => Unit): Unit = {
+  var i = 0
+  while (i < times) {
+    op
+    i += 1
+  }
+}
+
+inline def accumulate[T: ClassTag](size: Int)(inline op: => T): IArray[T] = {
```
sjrd, line 13: This looks like it's `IArray.fill(size)(op)`.

sjrd (commented):
This is for everything except the last commit (Scala 2 members)

jvm/src/test/scala/TypeSuite.scala
```diff
@@ -0,0 +1,34 @@
...
+
+  def assertMissingDeclaration(path: DeclarationPath)(using BaseContext): Unit =
+    absent(path)
+
+  def forceAbsentSymbol(path: DeclarationPath)(action: BaseContext ?=> Symbol)(using BaseContext): Symbol =
+    assertMissingDeclaration(path)
+    val found = action
+    val didForce = resolve(path)
+    assert(found eq didForce)
+    found
```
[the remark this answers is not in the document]
sjrd: It's to change this in a follow-up PR. Otherwise it's probably going to require re-doing a bunch of changes across several commits.

project/ClasspathSupport.scala
```diff
@@ -0,0 +1,6 @@
+import sbt._
+import sbt.Keys._
+
+object ClasspathSupport {
+  val LibraryClasspath = Configuration.of("ClasspathSupport.LibraryClasspath", "ClasspathSupport.LibraryClasspath")
```
sjrd, line 5: Any reason that this is not a top-level `val` in `build.sbt`?

shared/src/main/scala/tastyquery/ast/Symbols.scala
```diff
@@ -29,18 +29,35 @@ object Symbols {
     }
     final def tree: Tree = myTree
 
-    override def toString: String = s"symbol[$name]"
+    final def outer: Symbol = rawowner match {
+      case owner: Symbol => owner
+      case _             => assert(false, s"cannot access outer, $this was not declared within any scope")
```
sjrd, line 34:
```suggestion
      case null          => assert(false, s"cannot access outer, $this was not declared within any scope")
```
so that we can trust that all cases are covered through exhaustivity checking. Same below.

sjrd (commented):
And that's for the last commit. It looks like the bulk of it was copied from an existing pickle reader so I didn't review too carefully inside that area.

shared/src/main/scala/tastyquery/util/package.scala
```diff
@@ -0,0 +1,21 @@
+package tastyquery.util
+
+import scala.reflect.ClassTag
+
+inline def loop(times: Int)(inline op: => Unit): Unit = {
```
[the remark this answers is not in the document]
bishabosha: everywhere possible I am trying to avoid making function objects

project/ClasspathSupport.scala
bishabosha: I copy pasted this from scala 2 where I did it this way for the tasty-tests, I can move it to build.sbt

jvm/src/test/scala/TypeSuite.scala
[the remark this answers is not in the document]
bishabosha: I disagree, I would say the whole point of this test suite is to evaluate specifically the `resolveToSymbol` method. The end user would go through a different api that would produce trees where all types and symbols have been forced allready, with a read-only context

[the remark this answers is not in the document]
sjrd: Hum, I see. OK that makes sense. But it means that those internal APIs that are tested here should be package-private. Overall we'll need a future pass on the API to lock it down dramatically. There are plenty of stuff (also already on `main`) that is public and shouldn't be.

shared/src/main/scala/tastyquery/Contexts.scala
```diff
@@ -37,7 +37,25 @@ object Contexts {
...
       new FileContext(defn, defn.RootPackage, filename, classloader)
 
-    def createPackageSymbolIfNew(name: TermName, owner: Symbol): PackageClassSymbol = {
+    def createClassSymbolIfNew(name: Name, owner: DeclaringSymbol): ClassSymbol =
+      owner.getDecl(name) match {
+        case None =>
+          val cls = ClassSymbolFactory.createSymbol(name, owner)
+          owner.addDecl(cls)
+          cls
+        case Some(clazz) => ClassSymbolFactory.castSymbol(clazz)
```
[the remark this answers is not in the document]
bishabosha: In this case we need to rethink the way that classpath scanning works: currently we only traverse tasty files (which may make top level classes, e.g. `Foo` in say package `example`), then when we need the symbol for a Java dependency in `example` package, we then scan the classpath and create top level symbols - which will see `example.Foo` for the second time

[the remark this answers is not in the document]
bishabosha: I had avoided doing full classpath scanning from the beginning, because full dependency symbols would be unnecessary if we have a mode where someone just wants to see trees, with no semantic information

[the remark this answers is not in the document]
sjrd: > I had avoided doing full classpath scanning from the beginning, because full dependency symbols would be unnecessary if we have a mode where someone just wants to see trees

That sounds like premature optimization to me. *Scanning* a package in a classpath for top-level entries should never be so expensive that we would gain something substantial by only focusing on the ones with a `.tasty` file attached.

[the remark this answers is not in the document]
bishabosha: so to make this change, we will scan top level symbols first, then TreeUnpickler will need to be rewritten to make creating a classymbol a hard error when the symbol already exists - and then we will need to create a TreeUnpickler with the top level symbols in its constructor

shared/src/main/scala/tastyquery/util/package.scala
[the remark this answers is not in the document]
sjrd: OK. I don't think this is actually significant (we are creating lots of *other* objects anyway), but we can keep it as is for now.

shared/src/main/scala/tastyquery/ast/Symbols.scala
bishabosha: this is caused by Scalafmt, the line goes beyond the limit otherwise

jvm/src/main/scala/Main.scala
bishabosha: made the decision to load top level package symbols before unpickling in a follow up PR

build.sbt
```diff
@@ -18,6 +18,9 @@ lazy val tastyQuery =
       libraryDependencies += "org.scalameta" %% "munit" % "0.7.29" % Test,
       testFrameworks += new TestFramework("munit.Framework")
     )
+    .settings(
+      scalacOptions += "-Yexplicit-nulls"
```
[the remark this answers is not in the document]
bishabosha: removed Universe

jvm/src/test/scala/SymbolSuite.scala
bishabosha: Todo: make API's private

shared/src/main/scala/tastyquery/reader/TastyUnpickler.scala
bishabosha: got rid of the implicit stuff and renamed to `unsafe.asByteArray`

bishabosha:
all comments addressed

sjrd (approved):

PR #35, opened 2022-03-10
create top level classes before reading tasty

bishabosha:
- assert that symbols can only be defined once.
- user must obtain top-level class symbol before
  reading tasty - to obtain top-level class they
  must first scan classpath.
- when scanning for a top-level class, create
  symbols for all sibling classes in the same
  package.
- only complete the members of a top-level class
  on demand.
- same entry point to complete members both top-level
  tasty and java/scala2 class files.
- adds example where a top-level class will be seen twice.

### DIFF
--- a/jvm/src/test/scala/tastyquery/SymbolSuite.scala
+++ b/jvm/src/test/scala/tastyquery/SymbolSuite.scala
@@ -62,13 +62,6 @@ class SymbolSuite extends BaseUnpicklingSuite(withClasses = false, withStdLib = 
     )
   }
 
-  test("fail-on-symbolic-package:it-should-be-missing") {
-    intercept[MissingTopLevelDecl] {
-      getUnpicklingContext(name"symbolic_>>" / tname"#::") // this will fail, we can't find a symbolic package
-    }
-    val ctx = getUnpicklingContext(name"symbolic_$$greater$$greater" / tname"#::") // can only find encoded version
-  }
-
   test("basic-symbol-structure-nested") {
     val InNestedPackage = `simple_trees.nested` / tname"InNestedPackage"
     val ctx = getUnpicklingContext(InNestedPackage)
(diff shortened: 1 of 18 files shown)


shared/src/main/scala/tastyquery/Contexts.scala
```diff
@@ -19,6 +21,12 @@ object Contexts {
   transparent inline def clsCtx(using clsCtx: ClassContext): ClassContext = clsCtx
   transparent inline def defn(using baseCtx: BaseContext): baseCtx.defn.type = baseCtx.defn
 
+  object permissions:
+    opaque type InitAll = Unit
```
sjrd, line 25: It seems you did it again: creating an artificial permission to "ensure" that one specific method is only called where you have the "permission" to do so, but in fact "all" (i.e., one in this case) places where you call the method give themselves the permission immediately before calling it. This is only providing the illusion of safety, without making anything any safer. A permission system like that works if it makes sense for the permissions to be transitively passed through, and where the only way to create the permission is with something that *handles* the thing it permits. (like a `try/catch` creates the permission to `throw` within the `try`).

There is no need for this. Give `initPackages` a slightly scarier name like `initAllPackagesFromBaseContext`, and you will achieve the same level of "safety".

sjrd (changes requested):

shared/src/main/scala/tastyquery/Contexts.scala
bishabosha: my intention with this was just to make `initPackages` only callable from `Contexts`, (basically simulating an non-existent access control modifier) rather than global protection against a sub process eventually calling it - I will remove it though

jvm/src/test/scala/tastyquery/SymbolSuite.scala
```diff
@@ -63,13 +62,6 @@ class SymbolSuite extends BaseUnpicklingSuite(withClasses = false, withStdLib =
     )
   }
 
-  test("fail-on-symbolic-package:it-should-be-missing") {
```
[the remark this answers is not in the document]
bishabosha: done

shared/src/main/scala/tastyquery/Contexts.scala
```diff
@@ -141,47 +182,74 @@ object Contexts {
...
+
+      def mkSymbol(name: Name, owner: Symbol): T =
+        if owner == classRoot.enclosingDecl then
+          owner.lookup(name) match
+            case Some(sym) =>
+              if sym == classRoot && factory == ClassSymbolFactory then factory.castSymbol(classRoot)
+              else if sym.isModuleClassRoot && factory == ClassSymbolFactory then
+                factory.castSymbol(sym) // module class root
+              else if sym.isModuleRoot && factory == RegularSymbolFactory then factory.castSymbol(sym) // module root
+              else throw ExistingDefinitionException(owner, name, explanation = s"existing symbol: $sym")
```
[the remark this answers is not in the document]
bishabosha: done

shared/src/main/scala/tastyquery/Contexts.scala
bishabosha: removed

PR #43, opened 2022-03-20
Add positional information to trees

civp:
Adds `span` as a parameter of trees

Adds tests for `span`

Supports most of the subtrees except `Self`


sjrd (changes requested):
Two very high level comments to start.

build.sbt
```diff
@@ -28,6 +28,7 @@ lazy val tastyQuery =
     .settings(name := "tasty-query", version := "0.1-SNAPSHOT")
     .settings(
       libraryDependencies += "org.scala-lang" %% "tasty-core" % usedTastyRelease, // TODO: publish for JS or shade?
+      libraryDependencies += "org.scala-lang" %% "scala3-compiler" % "3.1.0",
```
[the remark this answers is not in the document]
chiefpi: Very useful information. Good to know this project is to decouple from the compiler.

shared/src/main/scala/tastyquery/ast/Trees.scala
```diff
@@ -17,7 +19,7 @@ object Trees {
     def unapply(e: TypeComputationError): Option[Tree] = Some(e.tree)
   }
 
-  abstract class Tree {
+  abstract class Tree(val span: Span) {
```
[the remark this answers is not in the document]
chiefpi: To be honest, I think `Span` is better because it implies that it consists of a start and an end, while `Position` sounds like it's a single point of position, like the `Coord` in dotc. If I understand correctly, we won't keep `Coord` in the code?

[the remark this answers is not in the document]
sjrd: Hum. Maybe you're right. Let's keep `span` for now.

build.sbt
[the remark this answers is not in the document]
chiefpi: It's removed now.

shared/src/main/scala/tastyquery/reader/PositionUnpickler.scala
```diff
@@ -0,0 +1,76 @@
...
+import dotty.tools.tasty.{TastyFormat, TastyBuffer, TastyReader}
+import TastyFormat.SOURCE
+import TastyBuffer.{Addr, NameRef}
+
+/** Unpickler for tree positions */
+class PositionUnpickler(reader: TastyReader, nameAtRef: NameRef => TermName) {
+  import reader.*
+
+  private var myLineSizes: Array[Int] = _
+  private var mySpans: util.HashMap[Addr, Span] = _
```
bishabosha, line 19: I think it is fine to use standard hash maps from the std library for now, if we identify the need we can optimise later

shared/src/main/scala/tastyquery/reader/TreeUnpickler.scala
```diff
@@ -428,200 +455,219 @@ class TreeUnpickler(protected val reader: TastyReader, nameAtRef: NameTable) {
...
+        SeqLiteral(reader.until(end)(readTerm), elemType)(curSpan)
+      case WHILE =>
+        reader.readByte()
+        reader.readEnd()
+        While(readTerm, readTerm)(curSpan)
+      case RETURN =>
+        reader.readByte()
+        val end = reader.readEnd()
+        val from = readSymRef
+        val idSpan = span
```
bishabosha, line 611: in Scala there is no labelled return so this should be an empty span

jvm/src/test/scala/tastyquery/PositionSuite.scala
```diff
@@ -0,0 +1,290 @@
...
+      collectCode[Lambda](tree, code),
+      // NoSpan is the expected behavior
+      Nil
+    )
+  }
+
+  test("return") {
+    val (tree, code) = unpickleWithCode(simple_trees / tname"Return")
+    assertEquals(collectCode[Return](tree, code), List("return 1"))
+    assertEquals(collectCode[Ident](tree, code), List("1"))
```
bishabosha, line 175: I'm not sure what this is, but if it is the Ident of the Return then the Id should have no span

jvm/src/test/scala/tastyquery/PositionSuite.scala
```diff
@@ -0,0 +1,290 @@
...
+    )
+  }
+
+  test("super") {
+    val (tree, code) = unpickleWithCode(simple_trees / tname"Super")
+    assertEquals(collectCode[Super](tree, code), List("super", "super[Base]"))
+  }
+
+  test("self".ignore) {
+    // ignore because the span of Self is impossible to construct
```
bishabosha, line 201: I looked at this file and it seems the span of the self def is the span of the identifier of the ValDef, and also it is a zero-length span unless the user explicitly gives a type, like for `TraitWithSelf`

bishabosha: it's worth using the main method of `dotty.tools.dotc.core.tasty.TastyPrinter` to debug tasty files

shared/src/main/scala/tastyquery/reader/TastyUnpickler.scala
```diff
@@ -45,6 +50,9 @@ class TastyUnpickler(reader: TastyReader) {
     this(new TastyReader(unsafe.asByteArray(bytes)))
 
   private val sectionReader = new mutable.HashMap[String, TastyReader]
+
+  /** Reserved for PositionUnpickler */
+  private val sectionReaderCloned = new mutable.HashMap[String, TastyReader]
```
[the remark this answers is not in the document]
chiefpi: I'm sorry I forgot to delete it.

shared/src/main/scala/tastyquery/reader/TreeUnpickler.scala
```diff
@@ -388,14 +412,17 @@ class TreeUnpickler(protected val reader: TastyReader, nameAtRef: NameTable) {
     if (reader.nextByte != SELFDEF) {
       reusable.EmptyValDef
     } else {
+      val curSpan = span
       reader.readByte()
       val name = readName
+      // TODO: Add span to TypeTree
```
bishabosha, line 418: so I think the correct span the type tree's span

shared/src/main/scala/tastyquery/reader/TreeUnpickler.scala
```diff
@@ -428,200 +455,219 @@ class TreeUnpickler(protected val reader: TastyReader, nameAtRef: NameTable) {
...
-    case SHAREDterm =>
-      reader.readByte()
-      forkAt(reader.readAddr()).readTerm
+        NamedArg(readName, readTerm)(curSpan)
+      case TYPEAPPLY =>
+        reader.readByte()
+        val end = reader.readEnd()
+        val fnSpan = span
+        val fn = readTerm
+        val paramsSpan = span
```
bishabosha, line 485: this would only get the span for the first type parameter, (all type trees have spans as well btw)

jvm/src/test/scala/tastyquery/PositionSuite.scala
```diff
@@ -0,0 +1,290 @@
...
+        "type AbstractWithBounds >: Null <: AnyRef",
+        "opaque type Opaque = Int",
+        "opaque type OpaqueWithBounds >: Null <: AnyRef = Null"
+      )
+    )
+  }
+
+  test("type-apply") {
+    val (tree, code) = unpickleWithCode(simple_trees / tname"TypeApply")
+    assertEquals(collectCode[TypeApply](tree, code), List("Seq"))
```
bishabosha, line 281: would be good to get a test with explicit type parameters also

shared/src/main/scala/tastyquery/reader/PositionUnpickler.scala
chiefpi: Done

shared/src/main/scala/tastyquery/reader/TastyUnpickler.scala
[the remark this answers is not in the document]
chiefpi: Now it's deleted :)

shared/src/main/scala/tastyquery/reader/TreeUnpickler.scala
chiefpi: Good to know. I'll add spans to `TypeTree`.

shared/src/main/scala/tastyquery/reader/TreeUnpickler.scala
chiefpi: You're right. It's fixed now.

chiefpi:
https://github.com/scalacenter/tasty-query/blob/7ba4447c45de2316fd36841a121e6f8dde3fb322/shared/src/main/scala/tastyquery/ast/Trees.scala#L95-L97

Some tests of spans are limited by the method `walkTypeTrees` because it assumes `TypeTree` can only be attached to `Tree`. But there're cases where `TypeTree` is a subtree of another `TypeTree`. For example, `MatchTypeTree` can be a subtree of `TypeLambdaTree`. In such cases, `walkTypeTrees` can't reach all the type trees. But I have to say the specification of the function is accurate.

shared/src/main/scala/tastyquery/reader/TreeUnpickler.scala
chiefpi: Fixed

shared/src/main/scala/tastyquery/reader/TreeUnpickler.scala
chiefpi: Fixed

jvm/src/test/scala/tastyquery/PositionSuite.scala
chiefpi: test cases are added

bishabosha:
this will need a rebase

jvm/src/test/scala/tastyquery/PositionSuite.scala
```diff
@@ -21,7 +23,11 @@ class PositionSuite extends BaseUnpicklingSuite(withClasses = false, withStdLib
   val imports = name"imports".singleton
 
   private def getCodePath(name: String): String =
-    s"${System.getProperty(ResourceCodeProperty)}/main/scala/${name.replace(".", "/")}.scala"
+    var Array(dir, filename) = name.split('.')
+    // A temporary workaround for path
+    // TODO: remove the workaround
+    if filename == "TraitWithSelf" then filename = "ClassWithSelf"
```
bishabosha, line 29: I guess we should make it so that there is only one top level class per test file

jvm/src/test/scala/tastyquery/PositionSuite.scala
```diff
@@ -0,0 +1,394 @@
...
+    val (tree, code) = unpickleWithCode(simple_trees / tname"VarargFunction")
+    assertEquals(collectCodeT[AnnotatedTypeTree](tree, code), List("Int*"))
+  }
+
+  test("match-type".ignore) {
+    val (tree, code) = unpickleWithCode(simple_trees / tname"MatchType")
+    assertEquals(collectCodeT[MatchTypeTree](tree, code), List(""))
+  }
+
+  test("type-tree-bind".ignore) {
```
bishabosha, line 368: why are these ignored?

shared/src/main/scala/tastyquery/reader/TreeUnpickler.scala
```diff
@@ -674,7 +740,8 @@ class TreeUnpickler(protected val reader: TastyReader, nameAtRef: NameTable) {
       case CaseDefFactory =>
         val pattern = readTerm
         val body = readTerm
-        CaseDef(pattern, reader.ifBefore(end)(readTerm, EmptyTree), body)
+        println(pattern)
+        CaseDef(pattern, reader.ifBefore(end)(readTerm, EmptyTree), body)(pattern.span.union(body.span))
```
bishabosha, line 744: `CASEDEF` nodes in a tasty file are associated with their own span, so we should use that

jvm/src/test/scala/tastyquery/PositionSuite.scala
civp: > https://github.com/scalacenter/tasty-query/blob/7ba4447c45de2316fd36841a121e6f8dde3fb322/shared/src/main/scala/tastyquery/ast/Trees.scala#L95-L97
> 
> Some tests of spans are limited by the method `walkTypeTrees` because it assumes `TypeTree` can only be attached to `Tree`. But there're cases where `TypeTree` is a subtree of another `TypeTree`. For example, `MatchTypeTree` can be a subtree of `TypeLambdaTree`. In such cases, `walkTypeTrees` can't reach all the type trees. But I have to say the specification of the function is accurate.

Because of the limitation of `walkTypeTrees` mentioned above

civp: I think it's easy to fix

shared/src/main/scala/tastyquery/reader/TreeUnpickler.scala
civp: Fixed!

jvm/src/test/scala/tastyquery/PositionSuite.scala
bishabosha: I see, I guess in a separate PR we can define walkTrees/TypeTrees on TypeTree and use that recursively in here

bishabosha (approved):
Fantastic, thank you very much for your contribution!

shared/src/main/scala/tastyquery/ast/Trees.scala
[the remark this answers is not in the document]
bishabosha: on this point, if we choose to have some span attached to definitions then `Coord` is how you link to things without trees, such as scala 2 symbols / java symbols

PR #44, opened 2022-03-24
Give types to all the simple things.

sjrd:
This requires synthetic declarations for `scala.Nothing` and `scala.Null`. They are added when we initialize the root context.

Applications do not receive types yet, because that will require `calculateType` to actually resolve symbols to get to functions' `MethodType`s.

### DIFF
--- a/shared/src/main/scala/tastyquery/ast/Constants.scala
+++ b/shared/src/main/scala/tastyquery/ast/Constants.scala
@@ -22,6 +22,20 @@ object Constants {
     import java.lang.Double.doubleToRawLongBits
     import java.lang.Float.floatToRawIntBits
 
+    def wideType: Type = tag match
+      case UnitTag    => Types.UnitType
+      case BooleanTag => Types.BooleanType
+      case CharTag    => Types.CharType
+      case ByteTag    => Types.ByteType
+      case ShortTag   => Types.ShortType
+      case IntTag     => Types.IntType
+      case LongTag    => Types.LongType
+      case FloatTag   => Types.FloatType
+      case DoubleTag  => Types.DoubleType
+      case StringTag  => Types.StringType
+      case NullTag    => Types.NullType
+      case ClazzTag   => Types.ClassTypeOf(typeValue)
+
     def isByteRange: Boolean = isIntRange && Byte.MinValue <= intValue && intValue <= Byte.MaxValue
     def isShortRange: Boolean = isIntRange && Short.MinValue <= intValue && intValue <= Short.MaxValue
     def isCharRange: Boolean = isIntRange && Char.MinValue <= intValue && intValue <= Char.MaxValue
--- a/shared/src/main/scala/tastyquery/ast/Trees.scala
+++ b/shared/src/main/scala/tastyquery/ast/Trees.scala
@@ -1,13 +1,10 @@
 package tastyquery.ast
 
 import tastyquery.ast.Constants.Constant
-import tastyquery.ast.Names.{Name, TermName, TypeName, nme}
-import tastyquery.ast.Types.{NoType, Type, TypeBounds, TypeRef}
+import tastyquery.ast.Names.*
+import tastyquery.ast.Types.*
 import tastyquery.ast.TypeTrees.*
-import tastyquery.ast.Symbols.{ClassSymbol, NoSymbol, PackageClassSymbol, RegularSymbol, Symbol}
-import tastyquery.ast.Names.SignedName
-import tastyquery.ast.Types.TermRef
-import tastyquery.ast.Types.PackageRef
+import tastyquery.ast.Symbols.*
 import tastyquery.util.syntax.chaining.given
 
 object Trees {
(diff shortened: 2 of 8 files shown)


shared/src/main/scala/tastyquery/ast/Trees.scala
```diff
@@ -259,22 +275,34 @@ object Trees {
   /**  @param meth   A reference to the method.
     *  @param tpt    Not an EmptyTree only if the lambda's type is a SAMtype rather than a function type.
     */
-  case class Lambda(meth: Tree, tpt: TypeTree) extends Tree
+  case class Lambda(meth: Tree, tpt: TypeTree) extends Tree {
+    override def calculateType: Type =
+      tpt.toType
```
bishabosha, line 280: this isn't right because `tpt` is only populated for SAM types, here is what dotty does:
```scala
def assignType(tree: untpd.Closure, meth: Tree, target: Tree)(using Context): Closure =
    tree.withType(
      if (target.isEmpty) meth.tpe.widen.toFunctionType(isJava = meth.symbol.is(JavaDefined), tree.env.length)
      else target.tpe)
```

bishabosha: `toFunctionType` is fairly complex

sjrd: Looks like this is another case where we will need to be able to resolve `meth` to a `MethodType`, so I stubbed it out for now.

shared/src/main/scala/tastyquery/ast/Trees.scala
```diff
@@ -324,9 +369,15 @@ object Trees {
     *
     * { bindings; expr }
     */
-  case class Inlined(expr: Tree, caller: TypeIdent, bindings: List[Tree]) extends Tree
+  case class Inlined(expr: Tree, caller: TypeIdent, bindings: List[Tree]) extends Tree {
+    override def calculateType: Type =
+      expr.tpe
```
bishabosha, line 374: todo: type avoidance on expr using the bindings

shared/src/main/scala/tastyquery/ast/TypeTrees.scala
```diff
@@ -35,45 +37,95 @@ object TypeTrees {
...
   /** qualifier#name */
-  case class SelectTypeTree(qualifier: TypeTree, name: TypeName) extends TypeTree
+  case class SelectTypeTree(qualifier: TypeTree, name: TypeName) extends TypeTree {
+    override protected def calculateType: Type =
+      TypeRef(qualifier.toType, name) // TODO Is this correct?
+  }
 
   /** qualifier.TypeName */
-  case class SelectTypeFromTerm(qualifier: Tree, name: TypeName) extends TypeTree
+  case class SelectTypeFromTerm(qualifier: Tree, name: TypeName) extends TypeTree {
```
bishabosha, line 78: this tree shouldn't really exists, in the tasty is corresponds to some path in a singleton type, so name should be `TermName`, and `calculateType` should be a `TermRef`

shared/src/main/scala/tastyquery/ast/TypeTrees.scala
```diff
@@ -35,45 +37,95 @@ object TypeTrees {
...
+
+  case class NamedTypeBoundsTree(name: TypeName, bounds: TypeBounds) extends TypeTree {
+    override protected def calculateType: Type =
+      NamedTypeBounds(name, bounds)
+  }
 
-  case class TypeLambdaTree(tparams: List[TypeParam], body: TypeTree) extends TypeTree
+  case class TypeLambdaTree(tparams: List[TypeParam], body: TypeTree) extends TypeTree {
+    override protected def calculateType: Type =
+      TypeLambda(tparams)(binders => body.toType) // TODO Shouldn't we use the binders here?
```
bishabosha, line 129: the binders parameter is just a hack for modifying the context when we see `TYPELAMBDAtype` - it can be ignored in favour of the `tparams` here.
Here is what dotty does:
```scala
def assignType(tree: untpd.LambdaTypeTree, tparamDefs: List[TypeDef], body: Tree)(using Context): LambdaTypeTree =
    tree.withType(HKTypeLambda.fromParams(tparamDefs.map(_.symbol.asType), body.tpe))
```

shared/src/main/scala/tastyquery/ast/Types.scala
```diff
@@ -25,11 +25,41 @@ object Types {
 
     def isOfClass(sym: Symbol)(using BaseContext): Boolean =
       this match {
-        case tpe: TermRef => tpe.underlying.isOfClass(sym)
-        case _            => this.isRef(sym)
+        case tpe: TermRef =>
+          tpe.underlying.isOfClass(sym)
+        case ConstantType(constant) =>
+          constant.wideType.isOfClass(sym)
```
bishabosha, line 31: `ConstantType.underlying` should delegate to `wideType`

shared/src/main/scala/tastyquery/ast/Constants.scala
```diff
@@ -22,6 +22,20 @@ object Constants {
...
+      case CharTag    => Types.CharType
+      case ByteTag    => Types.ByteType
+      case ShortTag   => Types.ShortType
+      case IntTag     => Types.IntType
+      case LongTag    => Types.LongType
+      case FloatTag   => Types.FloatType
+      case DoubleTag  => Types.DoubleType
+      case StringTag  => Types.StringType
+      case NullTag    => Types.NullType
+      case ClazzTag   => Types.ClazzType
```
bishabosha, line 37: this should be `AppliedType(java.lang.Class, this.typeValue :: Nil)`

bishabosha (approved):

PR #45, opened 2022-03-24
Make default symbol/decl lookup load things when necessary.

sjrd:
The previous behavior of finding decls without loading anything was renamed as `getDeclInternal`. It is used in a couple of internal locations where we really have no business loading stuff, because we are already within code that loads.

Otherwise, methods that lookup symbols require an implicit `BaseContext` so that they can load things when necessary.

This allows to remove the hacks in the TypeSuite to load symbols. They get loaded automatically.

### DIFF
--- a/jvm/src/test/scala/tastyquery/BaseUnpicklingSuite.scala
+++ b/jvm/src/test/scala/tastyquery/BaseUnpicklingSuite.scala
@@ -54,11 +54,8 @@ abstract class BaseUnpicklingSuite(withClasses: Boolean, withStdLib: Boolean, al
     (base, classRoot)
   }
 
-  /* TODO Currently this only resolves symbols that have already been loaded,
-   * but the intent is that it will load symbols when required.
-   */
   def resolve(path: DeclarationPath)(using BaseContext): Symbol =
-    followPathImpl(defn.RootPackage, path).fold(fail(_), identity)
+    summon[BaseContext].findSymbolFromRoot(path.toNameList)
 
   def assertSymbolExistsAndIsLoaded(path: DeclarationPath)(implicit ctx: BaseContext): Unit =
     followPathImpl(defn.RootPackage, path).fold(fail(_), _ => ())
@@ -87,7 +84,7 @@ abstract class BaseUnpicklingSuite(withClasses: Boolean, withStdLib: Boolean, al
         p.name.toTermName.select(s)
       case _ => next
     }
-    root.getDecl(sel) match {
+    root.getDeclInternal(sel) match {
       case Some(someSym) => Right(someSym)
       case _ => Left(s"No declaration for ${next.toDebugString} [${sel.toDebugString}] in ${root.toDebugString}")
     }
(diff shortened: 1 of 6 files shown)


sjrd:
Attention: the CI does not seem to be triggering at the moment.

shared/src/main/scala/tastyquery/ast/Symbols.scala
```diff
@@ -158,8 +169,24 @@ object Symbols {
 
     override lazy val thisType: Type = PackageRef(this)
 
+    override def getDecl(name: Name)(using BaseContext): Option[Symbol] =
+      super.getDecl(name).orElse {
```
bishabosha, line 173: minor but this is going to call `summon[BaseContext].classloader.scanClass(this)` unnecessarily

sjrd: Ah, I had not realized that `PackageClassSymbol` extends `ClassSymbol`. I changed it to directly call `getDeclInternal` instead.

jvm/src/test/scala/tastyquery/BaseUnpicklingSuite.scala
```diff
@@ -54,16 +54,19 @@ abstract class BaseUnpicklingSuite(withClasses: Boolean, withStdLib: Boolean, al
     (base, classRoot)
   }
 
-  def resolve(path: DeclarationPath)(implicit ctx: BaseContext): Symbol =
-    followPath(defn.RootPackage, path)
+  /* TODO Currently this only resolves symbols that have already been loaded,
+   * but the intent is that it will load symbols when required.
+   */
+  def resolve(path: DeclarationPath)(using BaseContext): Symbol =
```
bishabosha, line 60: this resolve won't work to find nested packages, because packages are entered with their fully qualified name

jvm/src/test/scala/tastyquery/BaseUnpicklingSuite.scala
```diff
@@ -54,16 +54,19 @@ abstract class BaseUnpicklingSuite(withClasses: Boolean, withStdLib: Boolean, al
     (base, classRoot)
   }
 
-  def resolve(path: DeclarationPath)(implicit ctx: BaseContext): Symbol =
-    followPath(defn.RootPackage, path)
+  /* TODO Currently this only resolves symbols that have already been loaded,
```
bishabosha, line 57: Which test prompted this comment?

jvm/src/test/scala/tastyquery/BaseUnpicklingSuite.scala
bishabosha: ok, you changed the definition of lookup

jvm/src/test/scala/tastyquery/BaseUnpicklingSuite.scala
sjrd: Oh this comment should be removed in the second commit, actually. I removed it now.

bishabosha (approved):

PR #49, opened 2022-04-01
Tree types need context to compute

bishabosha:
removes the "lazy" types `Application` and `TypeApplication`


sjrd:
The commit "extract BaseContext fields into ClassData" doesn't seem to belong here. It does not help/affect any of the subsequent commits in this PR. I also find it questionable. It does make the constructors simpler, but makes everything else more complex! Unless we have a compelling reason to do this (which again, doesn't seem to be in this PR), I don't think we should do this change.

Otherwise, LGTM.

bishabosha:
@sjrd I have removed the questionable commit

sjrd (approved):

PR #55, opened 2022-06-02
Add a motivation for tasty-query in the readme.

sjrd:


README.md
```diff
@@ -2,7 +2,217 @@
...
+### Preamble: TASTy and separate compilation
+
+When compiling a codebase, a Scala compiler combines the source files of a project with the "binaries" of its dependencies.
+Since the project source files can reference symbols (all kinds of definitions in a Scala program, like `val`s and `class`es) from the dependencies, the compiler builds a representation of the dependencies.
+
+In Scala 2, this representation is limited to public definitions and their API-level types (also protected and package-private, but for the intent of this explanation, we refer to all of that as "public").
+This information is stored in Scala 2 class files as "pickles".
+The bodies of methods are excluded from this treatment.
+
+In Scala 3, virtually everything about the source program is stored in so-called TASTy files: the complete AST, with all the bodies, and the types of everything.
```
julienrf, line 22:
```suggestion
In Scala 3, this representation is stored in so-called [TASTy files](https://docs.scala-lang.org/scala3/guides/tasty-overview.html). They contain the complete AST of the program, with all their public definitions _and their bodies_.
```

README.md
```diff
@@ -2,7 +2,217 @@
...
+Or to be precise, enough information is stored to be able to *reconstruct* everything (in particular, the types of subexpressions in method bodies).
+
+This representation is used to type-check (and further "elaborate") the project source files, clearly.
```
julienrf, line 25:
```suggestion
This representation is used to type-check (and further "elaborate") the project source files.
```

README.md
```diff
@@ -2,7 +2,217 @@
...
+
+* Symbol: the identity of any definition in a Scala or Java program.
+  Examples include class definitions, local or member val/def definitions.
+  There is only symbol per *definition*: multiple references to the same local val all refer to the same symbol.
+* Tree: an Abstract Syntax Tree (AST) or a subtree thereof.
+  In the context of tasty-query, they refer to TASTy trees, which carry their symbol and type information.
+* Type: a type in a Scala program, such as `Int`, `List[String]` or `Bar { type Foo <: Baz[String] }`.
+* Pickle: the public symbol and type information stored in Scala 2 binaries.
+* Elaboration: the process of turning a Scala source program into a typed AST with TASTy-level information.
+  Elaboration includes type inference, implicit resolution, overload resolution and macro expansion.
```
julienrf, line 38:
```suggestion
  Elaboration includes type inference, context arguments resolution, overload resolution and macro expansion.
```

README.md
```diff
@@ -2,7 +2,217 @@
...
+  It is the process that gives *semantics* (meaning) to a Scala program.
+  Because of the nature of Scala, it is impossible to only *typecheck* a source program; it must be fully *elaborated*.
+* TASTy: fully elaborated abstract syntax trees for Scala 3 programs.
+
+### Reading TASTy files
+
+TASTy files are complex beasts.
```
julienrf, line 45:
```suggestion
[TASTy files](https://docs.scala-lang.org/scala3/guides/tasty-overview.html) are complex beasts.
```

README.md
```diff
@@ -2,7 +2,217 @@
...
+They represent everything there is to know about the semantics of a Scala 3 program, and then some.
```
julienrf, line 46:
```suggestion
They represent everything there is to know about the semantics of a Scala 3 program.
```

README.md
```diff
@@ -2,7 +2,217 @@
...
+Whereas Java class files can be read in isolation, and processed to some extent without knowing the classpath context, it is virtually impossible to make any sense of TASTy files without a full classpath.
```
julienrf, line 47: I don’t understand this distinction. Why do we need a full classpath to “process to some extend” TASTy files? Reading the next sentences, it seems to be because the types of expression trees have to be reconstructed, and this is where you need the full classpath. Is that right? If this is the case, maybe consider making this link more explicit?

README.md
```diff
@@ -2,7 +2,217 @@
...
+
+* *Backward* compatibility: if something "works" with environment A, it also works with environment B.
+* *Forward* compatibility: if something works with environment B, it also works with environment A.
+
+By default, and otherwise specified, we refer to *backward* compatibility.
+
+In the Scala 2 ecosystem, there were two kinds of compatibility: *source* and *binary* compatibility.
+
+* Source compatibility: if a program source successfully compiles with environment A, then it also successfully compiles, with the same meaning, with environment B.
+* Binary compatibility: if a program's classfiles successfully link with environment A, then they also successfully link, with the same meaning, with environment B.
```
julienrf, line 69: Maybe that’s a detail, but on the JVM there is no real linking process (which is something that usually happens before run-time), so the definition is not easy to apply to the JVM setting.

julienrf: Unless you refer to issues that manifest at compile-time with errors like “Symbol xxx missing on the classpath”? But in general, I think binary compatibility issues manifest themselves at run-time with various forms of `LinkageError`s, as you mention further.

README.md
```diff
@@ -2,7 +2,217 @@
...
+
+Scala 3 adds a third kind of compatibility:
+
+* TASTy compatibility: if a program's TASTy files successfully link with environment A, then they also successfully link, with the same meaning, with environment B.
```
julienrf, line 73: I think it would be good to detail a little bit more what this linking step is, because it is not the same as the one used in the definition of binary compatibility (unless I’m wrong?).

README.md
```diff
@@ -2,7 +2,217 @@
...
+}
+```
+
+```scala
+// Test.Scala
+object Test {
+  def main(args: Array[String]): Unit = {
+    println(M.foo())
+  }
+}
```
julienrf, line 143:
```suggestion
@main def test(): Unit = {
  println(M.foo())
}
```

README.md
```diff
@@ -2,7 +2,217 @@
...
+```
+
+We first compile everything, and we even verify that it runs:
+
+```
+$ cs launch scalac:3.1.2 -- -d bin/ L.scala M.scala Test.scala
+$ cs launch scala:3.1.2 -- -cp bin/ Test
```
julienrf, line 150:
```suggestion
$ cs launch scala:3.1.2 -- -cp bin/ test
```

(I think. To be consistent with the other suggestion above)

README.md
```diff
@@ -2,7 +2,217 @@
...
+  |            ^^^^^^^
+  |            Found:    Seq[Int]
+  |            Required: Int
+  | This location contains code that was inlined from M.scala:3
+
+longer explanation available when compiling with `-explain`
+1 error found
+```
+
+This happens even though, from the point of view of `Test` the change of `L` is both source and binary compatible, because it is not TASTy-compatible.
```
julienrf, line 186: It could be helpful to detail a little bit what is causing the incompatibility, and why `inline` plays an important role here.

README.md
sjrd: There *is* linking on the JVM. It happens at run-time. That's why they're called `LinkageError`s? ;)
It's in the spec: https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-5.html#jvms-5.4

julienrf: OK that makes sense to use this wording then. But I suggest putting here the description of how linking errors manifest in the different platforms, instead of having it below.

README.md
sjrd: AFAICT, the *process* of resolving these things, including contextual abstractions, is still called "implicit resolution". At least, that's how the compiler calls it.

README.md
julienrf: OK

README.md
sjrd: The two following sentences are indeed explanations of this. I have reworded a bit to make it clearer.

julienrf (approved):
Thank you! :+1:

PR #60, opened 2022-06-10
Use Windows-friendly paths.

sjrd:
git crashes when trying to create a directory with '>' in it on Windows.


bishabosha (approved):
interesting to know windows has both encodings

PR #62, opened 2022-06-10
Some fixes in preparation for reading types from Scala 2 pickles

sjrd:
The test, even in its truncated form, needs the first two commits of this PR to pass.

### DIFF
--- a/shared/src/main/scala/tastyquery/Contexts.scala
+++ b/shared/src/main/scala/tastyquery/Contexts.scala
@@ -188,14 +188,10 @@ object Contexts {
 
     def moduleRoot: RegularSymbol = Contexts.moduleRoot(classRoot)(using this)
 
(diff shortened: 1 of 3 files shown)


shared/src/main/scala/tastyquery/Contexts.scala
```diff
@@ -188,14 +188,10 @@ object Contexts {
 
     def moduleRoot: RegularSymbol = Contexts.moduleRoot(classRoot)(using this)
 
-    def createSymbol[T <: Symbol](name: Name, factory: SymbolFactory[T], addToDecls: Boolean = false): T =
-      owner.getDeclInternal(name) match
```
bishabosha, line 192: is there a reason to get rid of this guard? I guess its for overloads

sjrd: Yes, it's for overloads.

bishabosha:
is this PR still relevant?

sjrd:
Superseded by #63.

PR #63, opened 2022-06-16
Unpickle Scala 2 types

sjrd:

PR #65, opened 2022-06-21
Make `Symbol.{declaredType,tree,declarations}` be auto-loading.

sjrd:
~~Based on #63.~~

### DIFF
--- a/jvm/src/test/scala/tastyquery/ReadTreeSuite.scala
+++ b/jvm/src/test/scala/tastyquery/ReadTreeSuite.scala
@@ -1,20 +1,20 @@
 package tastyquery
 
+import dotty.tools.tasty.TastyFormat.NameTags
+
 import munit.Location
 
 import tastyquery.Contexts
 import tastyquery.ast.Constants.{ClazzTag, Constant, IntTag, NullTag}
 import tastyquery.ast.Names.*
-import tastyquery.ast.Symbols.{NoSymbol, Symbol}
+import tastyquery.ast.Symbols.*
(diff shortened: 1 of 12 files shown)


jvm/src/test/scala/tastyquery/ReadTreeSuite.scala
```diff
@@ -463,7 +477,9 @@ class ReadTreeSuite extends BaseUnpicklingSuite(withClasses = false, withStdLib
     assert(containsSubtree(intFieldMatch)(clue(tree)))
   }
 
-  testUnpickle("object", simple_trees / tname"ScalaObject") { tree =>
+  testUnpickleTopLevel("object", simple_trees / tname"ScalaObject") { tree =>
+    // FIXME This test should not need `testUnpickleTopLevel`
```
bishabosha, line 481: I guess this is a limitation of requiring a TopLevelDeclPath - you can only inspect classes

sjrd: No it's a problem with the classloading. The top-level module symbols don't know how to trigger the unpickling for the file. There's even a comment:
https://github.com/scalacenter/tasty-query/blob/435a4f540c0340bdc83af5a6ada66d8f74ae705f/shared/src/main/scala/tastyquery/reader/classfiles/Classpaths.scala#L205

bishabosha: oh yeah I remember I did that now - ok we can look at this another time :)

bishabosha (approved):

PR #68, opened 2022-07-25
Explicitly wrap TypeBounds as WildcardTypeBounds in AppliedTypes.

sjrd:
TypeBounds of the form `>: L <: H` used in AppliedTypes actually mean wildcards of the form `? >: L <: H`. Previously, we carried around `Type | TypeBounds` in many places. Now, we explicitly wrap `TypeBounds` in a `WildcardTypeBounds` in places where they mean wildcards.

A corresponding change is done to `TypeTree`s.


bishabosha (approved):
very nice!

PR #70, opened 2022-07-29
WiP asSeenFrom

sjrd:
Well, this is what I have so far. Opening a draft PR for reference, and if @bishabosha wants to take over from here in the coming weeks. One existing test is still failing. Something to do with the initialization of `declaredType` for a `ClassSymbol` of an inner class. It's unclear why it would even want to see the `declaredType`, but that's what it is.


jvm/src/test/scala/tastyquery/TypeSuite.scala
```diff
@@ -575,4 +589,77 @@ class TypeSuite extends UnrestrictedUnpicklingSuite {
...
+    val PolySelect = resolve(name"simple_trees" / tname"PolySelect").asClass
+    val IntClass = resolve(name"scala" / tname"Int")
+
+    val Some(DefDef(_, _, _, body, _)) = PolySelect.lookup(name"testGetter").get.tree: @unchecked
+
+    val Select(qual, getterName) = body: @unchecked
+
+    assert(clue(qual.tpe).isApplied(_.isOfClass(GenClass), List(_.isOfClass(IntClass))))
+    assertEquals(getterName, name"getter")
+    body.tpe
```
bishabosha, line 617: is this line necessary? it is evaluated again in the next line

bishabosha: ok test behaves the same if I delete it

sjrd: That would be leftover from debugging. ;)

shared/src/main/scala/tastyquery/ast/Substituters.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  def subst(bounds: TypeBounds, from: List[Symbol], to: List[Type])(using Context): TypeBounds =
+    if from.isEmpty then bounds
+    else new SubstMap(from, to).apply(bounds)
+
+  final class SubstBindingMap(from: Binders, to: Binders)(using Context) extends DeepTypeMap:
+    def apply(tp: Type): Type =
+      tp match
+        case tp: BoundType =>
+          if (tp.binders eq from) tp.copyBoundType(to.asInstanceOf[tp.BindersType]) else tp
```
bishabosha, line 34: style - use new if syntax (across this file, maybe elsewhere too)

shared/src/main/scala/tastyquery/ast/Symbols.scala
```diff
@@ -64,6 +69,18 @@ object Symbols {
       if isFlagsInitialized then myFlags
       else throw IllegalStateException(s"flags of $this have not been initialized")
 
+    final def is(flag: Flag): Boolean =
```
bishabosha, line 72: seems like (ideally if not already possible) `export` should be able to cover this, even if its adapting an extension method on an opaque type to a normal method on the outer class.

shared/src/main/scala/tastyquery/ast/Symbols.scala
```diff
@@ -46,7 +50,8 @@ object Symbols {
 
     private[tastyquery] final def withDeclaredType(tpe: Type): this.type =
       val local = myDeclaredType
-      if local != null then throw new IllegalStateException(s"reassignment of declared type to $this")
+      if local != null /*&& !local.isInstanceOf[ClassType]*/ then // TODO Figure out why ClassType's are reassigned
```
bishabosha, line 53: this also seems like some leftover comment

shared/src/main/scala/tastyquery/ast/Symbols.scala
bishabosha: (other than flags being a def)

shared/src/main/scala/tastyquery/reader/classfiles/JavaSignatures.scala
```diff
@@ -44,7 +44,7 @@ object JavaSignatures:
               map.asInstanceOf[Map[TypeName, RegularSymbol]].get(tname) match
                 case Some(sym) => Some(TypeRef(NoPrefix, sym))
                 case None      => lookupTParam(member.outer)
-            case pt: Binders =>
+            case pt: TypeLambdaType =>
```
bishabosha, line 47: this could remain `Binders` if `BoundType` was made to extend `Type`, then these lookup methods could be made abstract in `Binders`

shared/src/main/scala/tastyquery/reader/TreeUnpickler.scala
```diff
@@ -852,13 +855,16 @@ class TreeUnpickler(
...
+        resultUnpickler.readType(using fileCtx.withEnclosingBinders(lambdaAddr, b))
+      )
     case PARAMtype =>
       reader.readByte()
       reader.readEnd()
       val lambdaAddr = reader.readAddr()
       val num = reader.readNat()
-      TypeParamRef(fileCtx.getEnclosingBinders(lambdaAddr), num)
+      //TypeParamRef(fileCtx.getEnclosingBinders(lambdaAddr), num)
+      fileCtx.getEnclosingBinders(lambdaAddr).asInstanceOf[TypeLambdaType].paramRefs(num)
```
bishabosha, line 867: same here, we can move `paramRefs` to `Binders` if we make `BoundType` a `Type`

shared/src/main/scala/tastyquery/ast/Types.scala
```diff
@@ -66,24 +68,91 @@ object Types {
 
   abstract class Type {
 
-    final def paramInfos: List[Type] = this match
+    /*final def paramInfos: List[Type] = this match
```
bishabosha, line 71: delete code here

shared/src/main/scala/tastyquery/ast/Types.scala
```diff
@@ -66,24 +68,91 @@ object Types {
...
       case tpe: MethodType => tpe.paramTypes
       case tpe             => Nil
 
     final def tparamBounds: List[TypeBounds] = this match
       case tpe: PolyType => tpe.paramTypeBounds
-      case tpe           => Nil
+      case tpe           => Nil*/
 
     final def tparamRefs: List[TypeParamRef] = this match
```
bishabosha, line 79: if you deleted the other methods above, why not this one as well

shared/src/main/scala/tastyquery/ast/Types.scala
```diff
@@ -66,24 +68,91 @@ object Types {
...
+      case tpe: TypeLambdaType => tpe.paramRefs
+      case tpe                 => Nil
+
+    /** The type parameters of this type are:
+      * For a ClassInfo type, the type parameters of its class.
+      * For a typeref referring to a class, the type parameters of the class.
+      * For a refinement type, the type parameters of its parent, dropping
+      * any type parameter that is-rebound by the refinement.
+      */
+    final def typeParams(using Context): List[ParamInfo.TypeParamInfo] =
```
bishabosha, line 89: this should probably be internal maybe

bishabosha: In dotty this looks like a utility for just computing type applications so I guess it should not be so visible

shared/src/main/scala/tastyquery/ast/Types.scala
```diff
@@ -66,24 +68,91 @@ object Types {
...
+        case self: TypeRef =>
+          val tsym = self.symbol
+          if (tsym.isClass) tsym.typeParams
+          else tsym.declaredType.typeParams
+        case self: AppliedType =>
+          if (self.tycon.typeSymbol.isClass) Nil
+          else self.superType.typeParams
+        case self: ClassType =>
+          self.cls.typeParams
+        case _: SingletonType | _: RefinedType =>
```
bishabosha, line 100: missing case for `TypeLambda` but that can be added when we test for it

shared/src/main/scala/tastyquery/ast/Types.scala
```diff
@@ -107,6 +176,93 @@ object Types {
       case tp: RefinedType   => tp.parent.widen
       case tp                => tp
 
+    final def widenIfUnstable(using Context): Type = this match
+      case tp: TermRef => tp
```
bishabosha, line 180: could be unstable value, or method

shared/src/main/scala/tastyquery/ast/Types.scala
```diff
@@ -173,6 +362,20 @@ object Types {
 
     /** The type to which this proxy forwards operations. */
     def underlying(using Context): Type
+
+    /** The closest supertype of this type.
+      *
+      * This is the same as `underlying`, except that
+      *   - instead of a TyperBounds type it returns its upper bound, and
```
bishabosha, line 369:
```suggestion
      *   - instead of a TypeBounds type it returns its upper bound, and
```

shared/src/main/scala/tastyquery/ast/Types.scala
```diff
@@ -207,6 +413,9 @@ object Types {
   }
 
   trait Symbolic {
+    final def symbol(using Context): Symbol =
```
bishabosha, line 416: what is the difference from `resolveToSymbol`?

shared/src/main/scala/tastyquery/ast/Types.scala
```diff
@@ -268,23 +477,26 @@ object Types {
...
         sym
       case name: Name =>
-        val prefixSym = prefix match {
+        def findPrefixSym(prefix: Type): Symbol = prefix match {
           case NoPrefix =>
             throw new SymbolLookupException(name, "reference by name to a local symbol")
           case t: TermRef =>
             val underlyingType = t.underlying
-            if (underlyingType == NoType)
+            /*if (underlyingType == NoType)
```
bishabosha, line 485: probably can delete this

shared/src/main/scala/tastyquery/ast/Types.scala
```diff
@@ -470,147 +784,448 @@ object Types {
...
+
+    def derivedAppliedType(tycon: Type, args: List[Type])(using Context): AppliedType =
+      if ((tycon eq this.tycon) && (args eq this.args)) this
+      else AppliedType(tycon, args)
+
+    def map(op: Type => Type)(using Context): AppliedType =
+      derivedAppliedType(op(tycon), mapConserve(args)(op))
   }
 
+  private def mapConserve[T <: AnyRef](xs: List[T])(op: T => T): List[T] =
```
bishabosha, line 827: blows stack

shared/src/main/scala/tastyquery/ast/Types.scala
```diff
@@ -470,147 +784,448 @@ object Types {
...
 
-  case object NoType extends GroundType
+    def derivedAndType(first: Type, second: Type)(using Context): Type =
+      if ((first eq this.first) && (second eq this.second)) this
+      else AndType.make(first, second)
+  }
+
+  object AndType {
+    def make(first: Type, second: Type)(using Context): Type =
+      if (first eq second) || (second eq AnyType) then first
```
bishabosha, line 1211: `AnyType` is not constant

shared/src/main/scala/tastyquery/ast/Types.scala
bishabosha: Actually, List now has a mapConserve method in the std library - should probably use this

shared/src/main/scala/tastyquery/ast/Symbols.scala
sjrd: `flags` is a `def` so it's not even stable. We can't export from it, regardless of whether it is extension methods or not.

sjrd:
I believe I have addressed all your comments so far.

shared/src/main/scala/tastyquery/ast/Types.scala
```diff
@@ -948,6 +906,9 @@ object Types {
 
     def companion: LambdaTypeCompanion[TermName, Type, MethodType] = MethodType
 
+    def findMember(name: Name, pre: Type)(using Context): Symbol =
```
bishabosha, line 909: possible improvement: `MethodOrPoly` super trait with this implementation

shared/src/main/scala/tastyquery/ast/Types.scala
```diff
@@ -992,10 +992,14 @@ object Types {
...
   end PolyType
 
-  /** Encapsulates the binders associated with a TypeParamRef. */
+  /** Encapsulates the binders associated with a ParamRef. */
   sealed trait Binders
 
-  sealed trait BoundType:
+  sealed trait TypeBinders extends Binders:
+    def paramRefs: List[TypeParamRef]
+    def lookupRef(name: TypeName): Option[Type]
```
bishabosha, line 1000: this could still be in `Binders` because it just returns type

bishabosha (approved):

PR #73, opened 2022-08-26
Fix nested package lookup

### LINKED ISSUE #71: Nested packages can't be found from their parent packages
`lang` cannot be found as a member of `java`. Test case:
```scala
    println(resolve(name"java" / name"lang" / tname"String")) // OK: symbol[class String]
    val javaPackage = resolve(name"java".singleton).asInstanceOf[PackageClassSymbol]
    println(javaPackage) // OK: symbol[package java]
    println(javaPackage.getDecl(name"lang")) // Not OK: None
    println(TypeRef(TermRef(PackageRef(javaPackage), name"lang"), tname"String").symbol) // Not OK: throws
```

bishabosha:
store packages in decls by their short name.

remove the `lookup` method, migration is to cast to `DeclaringSymbol` and call `getDecl`

fixes #71


sjrd (approved):
Nice simplifications :)

PR #75, opened 2022-08-29
unpickling top level classpath symbols can be done in any order

### LINKED ISSUE #72: Module classes cannot be found unless their companion class is loaded first
See the tests in `ReadTreeSuite` that use `testUnpickleTopLevel` and are marked with `FIXME`.

Maybe related: issues with `package$` package objects.

bishabosha:
with these changes you can force top-level module class, module value, or class in any order. The root class will be selected

fixes #72

### DIFF
--- a/shared/src/main/scala/tastyquery/ast/Symbols.scala
+++ b/shared/src/main/scala/tastyquery/ast/Symbols.scala
@@ -9,7 +9,10 @@ import tastyquery.ast.Trees.{DefTree, Tree, DefDef}
 import tastyquery.ast.Flags.*
 import tastyquery.ast.Types.*
 import tastyquery.ast.Variances.*
+import tastyquery.Contexts
 import tastyquery.Contexts.{Context, ctx}
+
+import tastyquery.util.syntax.chaining.given
 
 import compiletime.codeOf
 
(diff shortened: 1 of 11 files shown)


shared/src/main/scala/tastyquery/ast/Symbols.scala
```diff
@@ -34,20 +37,56 @@ object Symbols {
...
+      */
+    final def companionClass(using Context): Option[ClassSymbol] = maybeOuter match
+      case scope: DeclaringSymbol =>
+        this match
+          case _: PackageClassSymbol => None
+          case cls: ClassSymbol =>
+            scope.getDecl(cls.name.toTypeName.companionName).collect { case sym: ClassSymbol =>
+              sym
+            }
+          case obj: RegularSymbol if obj.isTerm =>
```
sjrd, line 76: This is too broad. With a case like
```
class Foo {
  class Bar
  def Bar: Int = 5
}
```
it would link `def Bar` to `class Bar`.

We should only allow modules here.

shared/src/main/scala/tastyquery/ast/Symbols.scala
```diff
@@ -196,6 +242,20 @@ object Symbols {
     private val myDeclarations: mutable.HashMap[Name, mutable.HashSet[Symbol]] =
       mutable.HashMap[Name, mutable.HashSet[Symbol]]()
 
+    /** Get the companion object of class definition, if it exists:
+      * - for `class C` => `object val C`
+      * - for `object class C[$]` => `object val C`
+      */
+    final def companionObject(using Context): Option[RegularSymbol] = maybeOuter match
```
sjrd, line 249: Could this be in `ClassSymbol` instead?

shared/src/main/scala/tastyquery/ast/Symbols.scala
```diff
@@ -296,19 +343,21 @@ object Symbols {
...
-          case _                                     => false
+        case pkg: PackageClassSymbol =>
+          ctx.classloader.scanPackage(pkg)
+          assert(initialised, s"could not initialize package $fullName")
+        case cls =>
+          def initRoot(root: ClassSymbol): Unit =
+            ctx.classloader.scanClass(root)
+            assert(Contexts.initialisedRoot(root), s"could not initialize root class $fullName")
+          require(owner.isPackage, s"inner class was not initialised by owner") // root classes only
+          val maybeRoot =
```
sjrd, line 354: This `val` is never used.

shared/src/main/scala/tastyquery/ast/Symbols.scala
```diff
@@ -369,32 +418,46 @@ object Symbols {
...
 
-  abstract class SymbolFactory[T <: Symbol] {
-    def createSymbol(name: Name, owner: Symbol): T
+  sealed abstract class SymbolFactory[T <: Symbol] {
+    type OwnerSym <: Symbol
+
+    final def createSymbol(name: Name, owner: OwnerSym)(using Context): T =
+      factory(name, owner).useWith(_.postCreate)
+
+    private[Symbols] def factory(name: Name, owner: OwnerSym): T
```
sjrd, line 429: `protected[this]` instead?

sjrd (changes requested):

shared/src/main/scala/tastyquery/ast/Symbols.scala
bishabosha: true, for now we don't have flags set up for modules - either I do that or temporarily require `scope` be a `PackageClassSymbol`

shared/src/main/scala/tastyquery/ast/Symbols.scala
bishabosha: done

shared/src/main/scala/tastyquery/ast/Symbols.scala
bishabosha: I went with my suggestion for now

shared/src/main/scala/tastyquery/ast/Symbols.scala
bishabosha: thanks!

shared/src/main/scala/tastyquery/ast/Symbols.scala
bishabosha: done

sjrd (approved):

PR #77, opened 2022-09-01
Only enter top level root symbols when reading the class/tasty file.

bishabosha:
differences:
- getDecl will now force a
  top level class' inner definitions,
  this is not observable from the public API,
  but some LazyLoadingSuite tests needed to
  change.
- getTopLevelTasty now accepts any top level
  class or module symbol.

add test case: assert no class for standalone object

experiment: replace `objclass"foo"` with `tname"foo" / obj`

### DIFF
--- a/jvm/src/test/scala/tastyquery/TypeSuite.scala
+++ b/jvm/src/test/scala/tastyquery/TypeSuite.scala
@@ -108,7 +108,7 @@ class TypeSuite extends UnrestrictedUnpicklingSuite {
   applyOverloadedTest("apply-overloaded-gen")("callB", name"simple_trees" / tname"OverloadedApply" / tname"Box")
   applyOverloadedTest("apply-overloaded-nestedObj")(
     "callC",
-    name"simple_trees" / tname"OverloadedApply" / objclass"Foo" / name"Bar"
+    name"simple_trees" / tname"OverloadedApply" / tname"Foo" / obj / name"Bar"
   )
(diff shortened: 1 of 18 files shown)


sjrd (changes requested):
Nice :) I only have minor comments.

jvm/src/test/scala/tastyquery/TypeSuite.scala
```diff
@@ -662,4 +662,9 @@ class TypeSuite extends UnrestrictedUnpicklingSuite {
     assert(clue(body.tpe).isOfClass(IntClass))
   }
 
+  testWithContext("BooleanRef") {
+    // val BooleanRef = resolve(name"scala" / name"runtime" / tname"BooleanRef").asClass
+    val scalaRuntime = resolve(name"scala" / name"runtime").asInstanceOf[DeclaringSymbol]
+    scalaRuntime.declarations
```
adpi2, line 668: Added this failing test to show that the reading of Java signatures is too eager.

adpi2: I also added the rt.jar in the `JavaTestPlatform` to run this test, but only when running in Java 8, which is why this test is not yet compatible with other Java versions.

To read Java 9+ class files, one could use something like this:
https://github.com/scalacenter/scala-debug-adapter/blob/0a25d3424bf5acfb4dcbebeada8fae980f4dc4bb/core/src/main/scala/ch/epfl/scala/debugadapter/ClassSystem.scala#L28-L36

adpi2: The problem is described in https://github.com/scalacenter/tasty-query/issues/80 and it should probably fixed in another PR.

PR #85, opened 2022-09-08
Test whether it also fails as is in a PR.

sjrd:


sjrd:
It was transient, apparently.

PR #86, opened 2022-09-08
Support RefinedTypes a little bit more.

sjrd:
At least, we don't crash when reading them. We can also read
recursive ThisType references to themselves.


adpi2 (approved):
Tout ceci est merveilleux

PR #87, opened 2022-09-08
Make Symbol.signature more robust and expose Symbol.signedName.

sjrd:
`signedName` returns a `SignedName` with the `signature` if there is a signature, i.e., if the symbol has a `MethodicType`. Otherwise, it returns the regular `name`.


adpi2 (approved):
Formidable

PR #90, opened 2022-09-09
Adjust the name of Scala 2 module classes.

sjrd:
In Scala 2 pickles, they just have a simple TypeName, but in tasty-query they must have the module class suffix.


adpi2 (approved):
Magnifique

PR #97, opened 2022-09-11
Read TASTy RECtype tag

adpi2:


shared/src/main/scala/tastyquery/reader/TreeUnpickler.scala
```diff
@@ -996,6 +996,9 @@ class TreeUnpickler(
         underlying
     case tag if (isConstantTag(tag)) =>
       ConstantType(readConstant)
+    case RECtype =>
+      reader.readByte()
+      readType
```
sjrd, line 1001: This should definitely interact with `recursiveTypeAtAddr` with "itself" before diving into `readType`. That means building a `RecType` with a lambda that passes itself. Similarly to how it's done in dotty.

sjrd (changes requested):

sjrd:
Superseded by #100.

PR #110, opened 2022-09-14
Separate FullyQualifiedName's out of the Name hierarchy.

sjrd:
There are two use cases for fully qualified names:

* in erased names for Signatures, and
* in pickled TASTy, for package definitions and references.

Package symbols used to have a fully qualified `name`, but that was inconsistent wrt. other symbols, whose `name`s are relative to their owner. Now, the `name` of package symbols is also a simple name relative to their parent packages.

### DIFF
--- a/shared/src/main/scala/tastyquery/Contexts.scala
+++ b/shared/src/main/scala/tastyquery/Contexts.scala
@@ -48,7 +48,7 @@ object Contexts {
         root.pkg
           .getDecl(root.rootName.toTypeName)
           .collect { case cls: ClassSymbol => cls }
-          .toRight(SymbolLookupException(root.fullName, s"perhaps it is not on the classpath"))
+          .toRight(SymbolLookupException(root.rootName, s"in ${root.pkg.fullName}; perhaps it is not on the classpath"))
       }
 
     /** Does there possibly exist a root for the given binary name. Does not force any classes covered by the name */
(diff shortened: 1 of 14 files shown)


shared/src/main/scala/tastyquery/Contexts.scala
```diff
@@ -122,31 +139,18 @@ object Contexts {
...
-              } else {
-                // create intermediate packages
-                val newOwner = createPackageSymbolIfNew(prefix, owner)
-                createPackageSymbolIfNew(name, newOwner)
-              }
-            case _ =>
-              throw IllegalArgumentException(s"Unexpected package name: $name")
-          }
+    def createPackageSymbolIfNew(fullyQualifiedName: FullyQualifiedName): PackageClassSymbol =
+      require(!fullyQualifiedName.path.isEmpty, s"Trying to create a package for an empty fully qualified name")
```
bishabosha, line 150: should this be a smart ctor for FQN

shared/src/main/scala/tastyquery/ast/Symbols.scala
```diff
@@ -181,21 +181,15 @@ object Symbols {
         .takeWhile(s => s.maybeOuter.exists)
         .forall(s => s.isPackage || s.isClass && s.name.toTypeName.wrapsObjectName)
 
-    private def nameWithPrefix(addPrefix: Symbol => Boolean): Name =
-      if isPackage then name
+    private def nameWithPrefix(addPrefix: Symbol => Boolean): FullyQualifiedName =
+      if isPackage && (isRoot || owner.isRoot) then FullyQualifiedName(name :: Nil)
```
bishabosha, line 185: I don't think we should be able to store the root package name in a fully qualified name, FQN is relative to the root

bishabosha: this should crash for the root package, or allow empty FQN

shared/src/main/scala/tastyquery/ast/Symbols.scala
```diff
@@ -204,6 +198,7 @@ object Symbols {
 
     final def isClass: Boolean = this.isInstanceOf[ClassSymbol]
     final def isPackage: Boolean = this.isInstanceOf[PackageClassSymbol]
+    final def isRoot: Boolean = isPackage && name == nme.RootName
```
bishabosha, line 201: also null owner? (`this.maybeOuter == NoSymbol`)

shared/src/main/scala/tastyquery/ast/Symbols.scala
bishabosha, line 185: still should fail here for `isRoot == true`

PR #111, opened 2022-09-14
Fix unpickling type whose bounds depend on itself

adpi2:
In `scala.collection.:+` there is a type parameter `C <: SeqOps[A, CC, C]`.

This is what happen when we read the type of  `C`:
```
read type C
  read symbol C
  store symbol C
  read declared type of symbol C
    read type C
    get symbol C from entries
    store type C
  set declared type of symbol C 
store type C // crash
```

### DIFF
--- a/shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
+++ b/shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
@@ -55,7 +55,7 @@ class PickleReader {
     val tOpt = entries(i).asInstanceOf[T | Null]
     if tOpt == null then {
       val res = pkl.unsafeFork(index(i))(op)
-      assert(entries(i) == null, entries(i))
+      assert(entries(i) == null || (entries(i) == res), entries(i))
       entries(i) = res
       res
     } else tOpt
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
```diff
@@ -55,7 +55,6 @@ class PickleReader {
     val tOpt = entries(i).asInstanceOf[T | Null]
     if tOpt == null then {
       val res = pkl.unsafeFork(index(i))(op)
-      assert(entries(i) == null, entries(i))
```
sjrd, line 58: It would still be good to check that we don't overwrite the entry with something *different*. Perhaps
```scala
assert(entries(i) == null || (entries(i) eq res), entries(i))
```

adpi2: `entries(i)` and `res` are equal but not reference equal

sjrd: Hum, OK, I guess we have to take it as is then. It's a bit of a shame, though.

sjrd (approved):

PR #115, opened 2022-09-15
First step of dealing with Scala 2 existential types.

sjrd:
We can deal with them when they are direct AppliedTypes. There are other cases that dotc handles but that we do not handle yet, notably RefinedTypes.

### DIFF
--- a/shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
+++ b/shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
@@ -145,7 +145,17 @@ class PickleReader {
     val nameref = pkl.readNat()
     val name0 = at(nameref)(readName())
     val name1 = name0.decode
+
+    assert(entries(storeInEntriesAt) == null, entries(storeInEntriesAt))
     val owner = readLocalSymbolRef()
+
+    /* In some situations, notably involving EXISTENTIALtpe, reading the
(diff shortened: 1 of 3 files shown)


shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
```diff
@@ -493,6 +502,76 @@ class PickleReader {
...
+          case info: TypeRef if isBound(info) =>
+            val info1 = info.symbol.info
+            assert(info1.derivesFrom(defn.SingletonClass))
+            RefinedType(parent1, name, info1.mapReduceAnd(removeSingleton)(_ & _))
+          case info =>
+            tp.derivedRefinedType(parent1, name, info)
+        }*/
+      case tp @ AppliedType(tycon, args) =>
+        val tycon1 = tycon // tycon.safeDealias
+        if (tycon1 ne tycon) elim(tycon1.appliedTo(args))
```
adpi2, line 556: How can `tycon1 ne tycon`?

sjrd: When we actually do the `tycon.safeDealias` which is currently commented out. I'll comment this branch as well; it's confusing.

adpi2 (approved):

PR #117, opened 2022-09-17
Fix #116: Correctly erase polymorphic opaque type aliases.

### LINKED ISSUE #116: IArray should not always be erased to java.lang.Object
Given this definition:
```scala
trait IArraySig {
  def from(): IArray[String]
}
```

This test fails:
```scala
testWithContext("iarray") {
  val IArraySig = resolve(name"simple_trees" / tname"IArraySig").asClass

  val from = IArraySig.getDecl(name"from").get
  assertIsSignedName(from.signedName, "from", "():java.lang.String[]")
}
```

```
Clues {
  name.sig.toString: String = "():java.lang.Object"
  signature: String = "():java.lang.String[]"
}
```

sjrd:
In particular, `opaque type IArray[+A] = Array[_ <: A]`.

Before that, we had to correctly integrate and instantiate higher-kinded `TypeLambda`s, similarly to how we already do for `PolyType`s.

### DIFF
--- a/shared/src/main/scala/tastyquery/ast/Types.scala
+++ b/shared/src/main/scala/tastyquery/ast/Types.scala
@@ -104,7 +104,7 @@ object Types {
                 // Fast path
                 ClassRef(tycon.symbol.asClass)
               case _ =>
-                preErase(tpe.superType)
+                preErase(tpe.translucentSuperType)
         case tpe: Symbolic =>
           tpe.resolveToSymbol match
             case cls: ClassSymbol =>
@@ -439,6 +439,18 @@ object Types {
       case WildcardTypeBounds(bounds) => bounds.high
       case st                         => st
     }
+
+    /** Same as superType, except for two differences:
+      *
+      *  - opaque types are treated as transparent aliases
+      *  - applied type are matchtype-reduced if possible
+      *
+      * Note: the reason to reduce match type aliases here and not in `superType`
+      * is that `superType` is context-independent and cached, whereas matchtype
+      * reduction depends on context and should not be cached (at least not without
+      * the very specific cache invalidation condition for matchtypes).
+      */
+    def translucentSuperType(using Context): Type = superType
 
     def findMember(name: Name, pre: Type)(using Context): Symbol =
       underlying.findMember(name, pre)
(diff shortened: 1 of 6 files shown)


shared/src/main/scala/tastyquery/ast/Types.scala
```diff
@@ -831,6 +843,13 @@ object Types {
         case tycon: TypeProxy                       => tycon.superType.applyIfParameterized(args)
         case _                                      => AnyType
 
+    override def translucentSuperType(using Context): Type = tycon match
+      case tycon: TypeRef if tycon.symbol.isOpaqueTypeAlias =>
+        tycon.translucentSuperType.applyIfParameterized(args)
```
bishabosha, line 848: not verified but this seems to throw some red flags for if the aliased type has different type parameters

shared/src/main/scala/tastyquery/ast/Types.scala
```diff
@@ -1068,30 +1087,32 @@ object Types {
     def paramNum: Int
 
   case class TypeLambda(paramNames: List[TypeName])(
-    @constructorOnly paramTypeBoundsExp: TypeLambda => List[TypeBounds],
-    @constructorOnly resultTypeExp: TypeLambda => Type
+    paramTypeBoundsExp: TypeLambda => List[TypeBounds],
```
bishabosha, line 1090: curious as to why this had to change when you have implemented lookahead in the treeunpickler

shared/src/main/scala/tastyquery/ast/Types.scala
sjrd: If it has different type parameters, the problem would be malformed. Wouldn't it?

shared/src/main/scala/tastyquery/ast/Types.scala
sjrd: It's not so much for the `TreeUnpickler` changes as for the `TypeLambdaTree.toType` changes. Since we must be able to execute `toType` without looking at other files.

shared/src/main/scala/tastyquery/ast/Types.scala
bishabosha: I was thinking of an some thing like ```opaque type Foo[A, B] = List[A]``` then wouldn't that apply two params to list, but since this is what dotty does I won't question

sjrd: No, when you have that definition, in TASTy it's actually a
```scala
opaque type Foo = [A, B] =>> List[A]
```
with a `TypeLambda`. So you apply the type lambda, which substitutes `A` and `B` for what you apply it to (discarding `B` in the process).

shared/src/main/scala/tastyquery/ast/Types.scala
sjrd: The external references can be read lazily. The problem is with `integrate`, which is too eager in trying to resolve aliases and what not.

To avoid this, we should have a "syntactic" version of type substitution that only looks at local symbols, and leaves everything else alone. But that's another story.

adpi2 (approved):
LGTM

PR #121, opened 2022-09-21
Add inheritance

bishabosha:
This does just enough to look up a symbol in a parent class

- parent types are now a list
- patch types of applied constructors
- patch signatures of constructor symbols
- add member lookup to class symbol

### DIFF
--- a/shared/src/main/scala/tastyquery/ast/Trees.scala
+++ b/shared/src/main/scala/tastyquery/ast/Trees.scala
@@ -347,13 +347,11 @@ object Trees {
       // in this case, check the method part to see if it is a constructor selection.
       // if it is, then the result type is the type of new.
       original match
-        case partial: MethodOrPoly => partial // Nothing to do here, it is partially applied.
+        case partial: MethodicType => partial // Nothing to do here, it is partially applied.
         case _ =>
           methPart(fun) match
             case Select(newObj @ New(_), SignedName(nme.Constructor, _, nme.Constructor)) =>
               newObj.tpe
-            // TODO: dotty indicates that there could possibly be cases for `super.<init>` and
-            // `this.<init>` internally inside a class
             case _ =>
               original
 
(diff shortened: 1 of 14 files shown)


shared/src/main/scala/tastyquery/ast/Trees.scala
```diff
@@ -317,7 +342,20 @@ object Trees {
...
+      // in this case, check the method part to see if it is a constructor selection.
+      // if it is, then the result type is the type of new.
+      original match
+        case partial: MethodOrPoly => partial // Nothing to do here, it is partially applied.
+        case _ =>
+          methPart(fun) match
+            case Select(newObj @ New(_), SignedName(nme.Constructor, _, nme.Constructor)) =>
+              newObj.tpe
+            // TODO: dotty indicates that there could possibly be cases for `super.<init>` and
+            // `this.<init>` internally inside a class
```
sjrd, line 356: Those should stay `Unit`, actually, so it should be fine.

shared/src/main/scala/tastyquery/ast/Trees.scala
```diff
@@ -317,7 +342,20 @@ object Trees {
           throw NonMethodReference(s"application of args ${args.mkString} to $tpe")
 
     override def calculateType(using Context): Type =
-      resolveMethodType(fun.tpe, args.map(_.tpe))
+      val original = resolveMethodType(fun.tpe, args.map(_.tpe))
+      // If the resolved method type is not methodic, then it is the final result type.
+      // in this case, check the method part to see if it is a constructor selection.
+      // if it is, then the result type is the type of new.
+      original match
+        case partial: MethodOrPoly => partial // Nothing to do here, it is partially applied.
```
sjrd, line 350: Why not `MethodicType`? It couldn't be an `ExprType` for constructors anyway.

sjrd (approved):
LGTM, but needs a rebase.

PR #123, opened 2022-09-23
Some refactorings

sjrd:


shared/src/main/scala/tastyquery/ast/Types.scala
```diff
@@ -810,11 +797,15 @@ object Types {
...
-    override def toString: String = s"PackageRef($fullyQualifiedName)"
+    def findMember(name: Name, pre: Type)(using Context): Symbol =
+      symbol.getDecl(name).getOrElse {
+        throw SymbolLookupException(name, s"Cannot find a member $name in $symbol")
+      }
   }
 
   object PackageRef {
-    def unapply(r: PackageRef): Option[FullyQualifiedName] = Some(r.fullyQualifiedName)
+    def apply(packageSym: PackageClassSymbol): PackageRef =
```
bishabosha, line 807: dont constructor proxies cover this?

bishabosha: no due to 

> 2. Constructor proxy apply methods are generated for a concrete class provided
the class has a companion object (which might have been generated in step 1), and
that companion object does not already define a member named apply.
Each generated apply method forwards to one constructor of the class. It has the same type and value parameters as the constructor.

sjrd: Not in this case, because there is an actual `apply` method generated for the `case class` primary constructor. And since there is an actual `apply`, the other constructors are not considered for constructor proxies at all.

bishabosha (approved):
looks great

PR #124, opened 2022-09-26
Make package symbols separate from ClassSymbol's.

sjrd:
Previously, `PackageClassSymbol` was a subclass of `ClassSymbol`. The notion of `DeclaringSymbol` was confused with `ClassSymbol`.

We now make these concepts entirely distinct:

* Rename `PackageClassSymbol` to `PackageSymbol`.
* Make `DeclaringSymbol` a `trait extends Symbol`.
* `PackageSymbol` extends `DeclaringSymbol` but not `ClassSymbol`.
* Move the logic for initialization to `DeclaringSymbol` and clearly name it as `ensureDeclsInitialized()`, since it is only ensuring that declarations are ready to be used; not that the symbol as a hole is ready.
* Remove `ClassSymbol.initialised`, since it is not used anymore.

### DIFF
--- a/shared/src/main/scala/tastyquery/ast/Symbols.scala
+++ b/shared/src/main/scala/tastyquery/ast/Symbols.scala
@@ -45,22 +45,12 @@ object Symbols {
     private var myTree: Option[DefTree] = None
     private var myPrivateWithin: Option[Symbol] = None
 
-    /** Forces the symbol to be initialized.
-      *
-      * This method should never be called from within the initialisation of
-      * this symbol.
-      *
-      * For symbols that are not root symbols, this has no effect.
-      */
-    private[tastyquery] def ensureInitialised()(using Context): Unit
-
     def withTree(t: DefTree): this.type =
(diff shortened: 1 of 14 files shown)


shared/src/main/scala/tastyquery/ast/Symbols.scala
```diff
@@ -379,13 +361,10 @@ object Symbols {
...
     final def companionObject(using Context): Option[RegularSymbol] = maybeOuter match
-      case scope: PackageClassSymbol =>
-        this match
-          case _: PackageClassSymbol => None
-          case cls: ClassSymbol =>
-            scope.getDecl(cls.name.toTypeName.sourceObjectName).collect { case sym: RegularSymbol =>
-              sym
-            }
+      case scope: PackageSymbol =>
+        scope.getDecl(this.name.toTypeName.sourceObjectName).collect { case sym: RegularSymbol =>
```
bishabosha, line 365: now we have flags I guess we can do decl based on flags?

bishabosha (approved):
incredible

PR #127, opened 2022-09-27
Remove Types.Symbolic. Deal with NamedType's and ThisType's separately.

sjrd:
We also standardize on `symbol` being the public accessor for the symbol of a `NamedType`, and use an private `computeSymbol()` as implementation detail.

### DIFF
--- a/shared/src/main/scala/tastyquery/ast/Types.scala
+++ b/shared/src/main/scala/tastyquery/ast/Types.scala
@@ -87,7 +87,7 @@ object Types {
             val List(targ) = targs: @unchecked
             arrayOf(targ).arrayOf()
           else arrayOf(tycon)
-        case tpe: Symbolic =>
+        case tpe: TypeRef =>
           val sym = tpe.symbol
           if sym.isClass then ClassRef(sym.asClass).arrayOf()
           else arrayOf(sym.declaredType)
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/tastyquery/ast/Types.scala
```diff
@@ -568,22 +563,27 @@ object Types {
...
+    private def computeSymbol()(using Context): Symbol = designator match
+      case sym: Symbol =>
         sym
+      case LookupIn(pre, name) =>
+        TermRef(pre, name).symbol
       case Scala2ExternalSymRef(owner, path) =>
-        val sym = path.foldLeft(owner) { (owner, name) =>
-          val cls = if owner.isTerm then owner.declaredType.asInstanceOf[Symbolic].symbol else owner
+        path.foldLeft(owner) { (owner, name) =>
+          val cls = if owner.isTerm then owner.declaredType.asInstanceOf[TypeRef].symbol else owner
```
bishabosha, line 582: maybe a comment to say that owner.isTerm can only be module value so its declared type will always be the TypeRef to its module class

sjrd: Yes, good point. I added a comment.

bishabosha (approved):
looks good

PR #129, opened 2022-09-27
Store the parents of a class in the ClassSymbol instead of the ClassInfo.

sjrd:
Unlike in dotc, we don't have different periods, so we will always have the same parent list for any given class.


bishabosha:
perhaps we will have an encoding where associated class info changes based on different versions of the classpath?

sjrd:
Perhaps, but we have no idea, and the same "perhaps" can be asked of virtually *everything* we do currently. We will add the necessary indirections in the future if we need them.

sjrd:
In fact, we will almost certainly need different *symbols* for different versions of any given class on different classpaths. Because they would come from different .class/.tasty files, and so would be created in two different instances of the Unpicklers.

bishabosha (approved):
Looks good overall

PR #130, opened 2022-09-27
Use a dedicated tree node SelfDef for self type definitions.

sjrd:
This way, we do not need to store a NoSymbol in the tree, and we do not have meaningless fields `rhs` and `symbol`.


bishabosha (approved):
Nice

PR #131, opened 2022-09-28
The Great Symbols Refactoring.

sjrd:
Previously, the hierarchy of `Symbol` was limited to `ClassSymbol`, `PackageSymbol` and `RegularSymbol`. `RegularSymbol`s could equally represent terms or (non-class) types.

In the latter case, their `declaredType` would contain a dubious encoding of the type definition. Notably, we used `WildcardTypeBounds` as a way to wrap `TypeBounds` in a `Type`, but that was a hack, and they had to be extracted everywhere.

Now, the hierarchy is more elaborate. In particular, we distinguish `TermSymbol`s, with a `declaredType: Type`, from the `TypeSymbol`s, which are either `ClassSymbol`s or `TypeSymbolWithBounds`, the latter being further classified and exposing `bounds: TypeBounds`.

Some aspects of `Type`s are refined to expose and manipulate the appropriate subtypes of `Symbol`s. In particular, `TermRef`s are guaranteed to point to `TermSymbol`s while `TypeRef`s point to `TypeSymbol`s.

### DIFF
--- a/shared/src/main/scala/tastyquery/ast/Symbols.scala
+++ b/shared/src/main/scala/tastyquery/ast/Symbols.scala
@@ -18,6 +18,53 @@ import tastyquery.util.syntax.chaining.given
 import compiletime.codeOf
 import tastyquery.reader.classfiles.Classpaths.Loader
 
+/** Symbols for all kinds of definitions in Scala programs.
+  *
+  * Every definition, like `class`es, `def`s, `type`s and type parameters, is
+  * associated with a `Symbol`. `Symbol`s are organized in a hierarchy,
+  * depending on what kind of definitions they represent.
+  *
+  * ```none
+  * Symbol
+  *  |
+  *  +- NoSymbol                        singleton instance used when there is no symbol
+  *  |
+  *  +- PackageSymbol                   any package, including the root package, the empty package, and nested packages
+  *  |
+  *  +- TermSymbol                      any term definition:
+  *  |                                  `val`, `var`, `def`, term param, term capture, `object` value
+  *  |
+  *  +- TypeSymbol                      any definition for a type
+  *     +- ClassSymbol                  definition for a `class`, `trait`, or the module class of an `object`
+  *     +- TypeSymbolWithBounds         any other kind of type: `type` definitions, type params, type captures
+  *        +- TypeMemberSymbol          `type` definition, further refined through its `typeDef`
+  *        +- TypeParamSymbol
+  *           +- ClassTypeParamSymbol   type parameter of a class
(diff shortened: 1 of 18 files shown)


shared/src/main/scala/tastyquery/ast/Types.scala
```diff
@@ -1249,6 +1281,12 @@ object Types {
       bounds.high.findMember(name, pre)
   }
 
+  enum TypeMemberDefinition:
```
bishabosha, line 1284: these don't need to be in types

sjrd: I moved it to `Symbols`.

I also added a big documentation comment on `Symbols`, and expanded the commit message.

shared/src/test/scala/tastyquery/TypeSuite.scala
```diff
@@ -718,15 +718,16 @@ class TypeSuite extends UnrestrictedUnpicklingSuite {
...
   }
 
   testWithContext("scala2-type-alias") {
-    val PredefString = resolve(name"scala" / tname"Predef" / obj / tname"String")
+    val PredefString = resolve(name"scala" / tname"Predef" / obj / tname"String").asType
     val JLString = resolve(name"java" / name"lang" / tname"String")
 
-    assert(PredefString.declaredType.isRef(JLString))
+    assert(clue(PredefString).isTypeAlias)
+    assert(clue(PredefString.asInstanceOf[TypeMemberSymbol].aliasedType).isRef(JLString))
```
sjrd, line 730: I made this test more straightforward.

shared/src/main/scala/tastyquery/ast/Symbols.scala
```diff
@@ -243,18 +243,128 @@ object Symbols {
...
+  object TermSymbol:
+    private[tastyquery] def create(name: TermName, owner: Symbol): TermSymbol =
+      owner.addDeclIfDeclaringSym(TermSymbol(name, owner))
+  end TermSymbol
+
+  sealed abstract class TypeSymbol protected (val name: TypeName, override val owner: Symbol) extends Symbol(owner):
+    type ThisNameType = TypeName
+
+    // Overridden in TypeMemberSymbol
+    def isTypeAlias(using Context): Boolean = false
```
bishabosha, line 275: I’d prefer a final method and pattern match

sjrd: Done.

bishabosha (approved):
Very nice

PR #132, opened 2022-10-02
Move the main method to an example project.

sjrd:
Along with the supporting methods that were in `tastyquery.api.*`.

We get rid of the "JS main" which was nonsensical.


bishabosha (approved):
lgtm

PR #135, opened 2022-10-03
Refactor Classpath to make it part of the public API

sjrd:


tasty-query/shared/src/main/scala/tastyquery/Classpaths.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  /** Contains tasty bytes.
+    *
+    * `binaryName` is the file name without the `.class` extension.
+    */
+  final class TastyData(val binaryName: String, val debugPath: String, val bytes: IArray[Byte]):
+    override def toString(): String = s"TastyData($binaryName, $debugPath)"
+
+  final class Classpath private (val packages: IArray[PackageData]):
+    def withFilter(binaryNames: List[String]): Classpath =
```
bishabosha, line 23: `Paths.rootClassName` should re-encode name components before calling this method

bishabosha (changes requested):
just the point about filtering the classpath

tasty-query/shared/src/main/scala/tastyquery/Classpaths.scala
sjrd: Done.

bishabosha (approved):

PR #136, opened 2022-10-03
add parents to scala 2 classes, erase scala 2 value classes

### LINKED ISSUE #119: Erasure of value class
Here is the test that fails:

```scala
testWithContext("scala.collection.ArrayOps") {
  val Predef = resolve(name"scala" / tname"Predef" / obj).asClass
  val genericArrayOps = Predef.getDecl(name"genericArrayOps").get
  assertIsSignedName(genericArrayOps.signedName, "genericArrayOps", "(1,java.lang.Object):java.lang.Object")
}
```

Notice that `ArrayOps[T]` is a value class of `Array[T]` and should be erased to `Object`.

bishabosha:
also standardise on paramRefss rather than paramSymss.

fixes #119

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Types.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Types.scala
@@ -13,6 +13,8 @@ import tastyquery.Names.*
 import tastyquery.Symbols.*
 import tastyquery.Trees.*
 
+import scala.annotation.targetName
+
 object Types {
   type Designator = Symbol | Name | LookupIn | Scala2ExternalSymRef
 
@@ -141,8 +143,8 @@ object Types {
     private def finishErase(typeRef: ErasedTypeRef)(using Context): ErasedTypeRef =
       def valueClass(cls: ClassSymbol): ErasedTypeRef =
         val ctor = cls.getDecl(nme.Constructor).get.asTerm
-        val List(Left(List(param))) = ctor.paramSymss.dropWhile(_.isRight): @unchecked
-        val paramType = param.declaredType
+        val List(Left(List(paramRef))) = ctor.paramRefss.dropWhile(_.isRight): @unchecked
+        val paramType = paramRef.underlying
         erase(paramType)
 
       typeRef match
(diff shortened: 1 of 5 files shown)


tasty-query/shared/src/main/scala/tastyquery/Types.scala
```diff
@@ -379,10 +389,14 @@ object Types {
...
-        case tpe: AppliedType => tpe.underlying.isRef(sym)
-        case tpe: ExprType    => tpe.underlying.isRef(sym)
-        case _                => false // todo: add ProxyType (need to fill in implementations of underlying)
+        case tpe: NamedType    => tpe.symbol == sym
+        case tpe: AppliedType  => tpe.underlying.isRef(sym)
+        case tpe: ExprType     => tpe.underlying.isRef(sym)
+        case tpe: TermParamRef => tpe.underlying.isRef(sym)
+        case tpe: TypeParamRef =>
+          val bounds = tpe.bounds
+          bounds.low.isRef(sym) || bounds.high.isRef(sym)
```
sjrd, line 398: Shouldn't this be an `&&`? Or only test `bounds.high`?

tasty-query/shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
```diff
@@ -210,15 +214,26 @@ class PickleReader {
             .asInstanceOf[DeclaringSymbol]
             .getModuleDeclInternal(name.toTermName.stripObjectSuffix)
           module.foreach(m => m.asTerm.withDeclaredType(cls.typeRef))
+        val tpe = readSymType()
         val typeParams = atNoCache(infoRef)(readTypeParams())
+        if !isRefinementClass(cls) then
+          val parentTypes = tpe match
+            case TempPolyType(tparams, restpe) =>
+              assert(tparams.corresponds(typeParams)(_ eq _)) // should reuse the class type params
+              restpe.asInstanceOf[TempClassInfoType].parentTypes
```
sjrd, line 223: Consider:
```suggestion
            case TempPolyType(tparams, restpe: TempClassInfoType) =>
              assert(tparams.corresponds(typeParams)(_ eq _)) // should reuse the class type params
              restpe.parentTypes
```

sjrd (changes requested):

tasty-query/shared/src/main/scala/tastyquery/Types.scala
bishabosha: ive changed to only check `high`

tasty-query/shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
bishabosha: done

sjrd (approved):

PR #137, opened 2022-10-03
Hide or remove helpers

sjrd:


tasty-query/shared/src/main/scala/tastyquery/reader/UTF8Utils.scala
```diff
@@ -0,0 +1,14 @@
+package tastyquery.reader
+
+import scala.annotation.targetName
+
+import java.nio.charset.StandardCharsets
+
+object UTF8Utils:
+  def decode(bytes: Array[Byte], offset: Int, len: Int): String =
+    new String(bytes.asInstanceOf[Array[Byte]], offset, len, StandardCharsets.UTF_8)
```
bishabosha, line 9:
```suggestion
    new String(bytes, offset, len, StandardCharsets.UTF_8)
```

tasty-query/shared/src/main/scala/tastyquery/reader/classfiles/ClassfileReader.scala
```diff
@@ -84,14 +79,11 @@ final class ClassfileReader private () {
         throw ReadException(s"Expected unforced UTF8 constant at index $idx")
     }
 
-    /** returns a new IArray with the decoded bytes, mutates `bytes` in-place, so `bytes` should not be shared
-      * after passing to this method.
-      */
+    /** Returns a new IArray with the decoded bytes. */
     private def unsafeDecodeSigBytes(bytes: IArray[Byte]): IArray[Byte] =
```
bishabosha, line 83: plus fix the callsite
```suggestion
    private def decodeSigBytes(bytes: IArray[Byte]): IArray[Byte] =
```

bishabosha (approved):

PR #141, opened 2022-10-04
Lock down any access to the `designator` of `NamedType`s.

sjrd:
The designator is not a stable value. Only `name` and `symbol` should be accessed from the outside.

To hide the `designator`, we have to make `TermRef` and `TypeRef` non-case classes.

For testing purposes, we expose a `private[tastyquery]` access to the designator with a scary name.

These changes also allow to make `LookupIn` and `Scala2ExternalSymRef` `private[tastyquery]`, which furthers hide implementation details of the library.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/TypeTrees.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/TypeTrees.scala
@@ -121,7 +121,7 @@ object TypeTrees {
       extends TypeTree(span)
       with DefTree(symbol) {
     override protected def calculateType(using Context): Type =
-      TermRef(NoType, symbol)
+      TypeRef(NoType, symbol)
 
     override final def withSpan(span: Span): TypeTreeBind = TypeTreeBind(name, body, symbol)(span)
   }
--- a/tasty-query/shared/src/main/scala/tastyquery/Types.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Types.scala
@@ -16,11 +16,9 @@ import tastyquery.Trees.*
 import scala.annotation.targetName
 
 object Types {
-  type Designator = Symbol | Name | LookupIn | Scala2ExternalSymRef
+  private[tastyquery] final case class LookupIn(pre: TypeRef, sel: SignedName)
 
-  case class LookupIn(pre: TypeRef, sel: SignedName)
-
-  case class Scala2ExternalSymRef(owner: Symbol, path: List[Name]) {
+  private[tastyquery] final case class Scala2ExternalSymRef(owner: Symbol, path: List[Name]) {
     val name = path.last
   }
 
@@ -542,9 +540,9 @@ object Types {
   }
 
   trait PathType extends TypeProxy with ValueType {
-    final def select(name: Name): NamedType =
-      if name.isTermName then TermRef(this, name)
-      else TypeRef(this, name)
+    final def select(name: Name): NamedType = name match
+      case name: TermName => TermRef(this, name)
+      case name: TypeName => TypeRef(this, name)
   }
 
   // ----- Type Proxies -------------------------------------------------
@@ -554,12 +552,17 @@ object Types {
 
     type ThisName <: Name
     type ThisSymbolType <: Symbol { type ThisNameType = ThisName }
+    protected[this] type ThisDesignatorType >: ThisSymbolType <: Symbol | Name | LookupIn | Scala2ExternalSymRef
 
     val prefix: Type
 
-    def designator: Designator
+    protected[this] def designator: ThisDesignatorType
 
-    protected def designator_=(d: Designator): Unit
+    protected[this] def storeComputedSymbol(sym: ThisSymbolType): Unit
+
+    // For tests
+    private[tastyquery] final def designatorInternal: Symbol | Name | LookupIn | Scala2ExternalSymRef =
+      designator
 
     private var myName: ThisName | Null = null
     private var mySymbol: ThisSymbolType | Null = null
@@ -596,7 +599,7 @@ object Types {
       else
         val computed = asThisSymbolType(computeSymbol())
         mySymbol = computed
-        designator = computed
+        storeComputedSymbol(computed)
         computed
 
     protected[this] def asThisSymbolType(sym: Symbol): ThisSymbolType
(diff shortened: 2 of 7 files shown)


tasty-query/shared/src/main/scala/tastyquery/TypeTrees.scala
```diff
@@ -121,7 +121,7 @@ object TypeTrees {
       extends TypeTree(span)
       with DefTree(symbol) {
     override protected def calculateType(using Context): Type =
-      TermRef(NoType, symbol)
+      TypeRef(NoType, symbol)
```
bishabosha, line 124: TypeTreeBind is for a match type right? so this change doesn't seem to make sense

tasty-query/shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
```diff
@@ -543,7 +545,7 @@ private[pickles] class PickleReader {
       case LITERALclass   => Constant(readTypeRef())
       case LITERALenum =>
         readMaybeExternalSymbolRef() match
-          case sym: Symbol                 => TermRef(NoPrefix, sym)
+          case sym: Symbol                 => TermRef(NoPrefix, sym.asTerm)
```
bishabosha, line 548: why not?
```suggestion
          case sym: TermSymbol                 => TermRef(NoPrefix, sym)
```

sjrd: Because we'd still need to deal with other `Symbol`s for the match to be exhaustive.

tasty-query/shared/src/main/scala/tastyquery/Types.scala
```diff
@@ -844,14 +859,17 @@ object Types {
...
-  case class TypeRef(override val prefix: Type, private var myDesignator: Designator) extends NamedType {
+  final class TypeRef private (val prefix: Type, myDesignator: TypeName | TypeSymbol | Scala2ExternalSymRef)
+      extends NamedType {
 
     type ThisName = TypeName
     type ThisSymbolType = TypeSymbol
+    type ThisDesignatorType = TypeName | TypeSymbol | Scala2ExternalSymRef
 
-    override def designator: Designator = myDesignator
+    override def designator: ThisDesignatorType = myDesignator
```
bishabosha, line 869: this `designator` is public, is this intentional?

tasty-query/shared/src/main/scala/tastyquery/Types.scala
```diff
@@ -768,17 +771,19 @@ object Types {
...
       extends NamedType
       with SingletonType
       with ImplicitRef {
 
     type ThisName = TermName
     type ThisSymbolType = TermSymbol
+    protected[this] type ThisDesignatorType = TermSymbol | TermName | LookupIn | Scala2ExternalSymRef
 
-    override def designator: Designator = myDesignator
+    override def designator: ThisDesignatorType = myDesignator
```
bishabosha, line 783: this `designator` is public, is this intentional?

tasty-query/shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
bishabosha: id recommend `: @unchecked` but up to you

bishabosha (changes requested):
I think the subclasses should have protected modifier on `designator`

tasty-query/shared/src/main/scala/tastyquery/Types.scala
```diff
@@ -844,14 +859,17 @@ object Types {
       new PackageRef(packageSym)
   }
 
-  case class TypeRef(override val prefix: Type, private var myDesignator: Designator) extends NamedType {
+  final class TypeRef private (val prefix: Type, myDesignator: TypeName | TypeSymbol | Scala2ExternalSymRef)
```
bishabosha, line 862: now `myDesignator` is immutable, I think it should still be nulled out or replaced by the type symbol when calculated - otherwise we retain the `Scala2ExternalSymRef` (or `LookupIn` for TermRef)

bishabosha: then the `designatorInternal` makes sense to have a scary name, now its just returning a stable field

tasty-query/shared/src/main/scala/tastyquery/Types.scala
sjrd: Ouch! No, definitely not intentional.

tasty-query/shared/src/main/scala/tastyquery/Types.scala
sjrd: OK, I reintroduced storing the symbol in the designator. This will also indirectly help with `toString`-based debugging.

tasty-query/shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
sjrd: I added an explicit `errorBadSignature` when it's not a `TermSymbol`.

tasty-query/shared/src/main/scala/tastyquery/Types.scala
```diff
@@ -596,7 +599,6 @@ object Types {
       else
         val computed = asThisSymbolType(computeSymbol())
         mySymbol = computed
-        designator = computed
```
bishabosha, line 599: Call `storeComputedSymbol` here maybe?

sjrd: Oh, yes. 🤦‍♂️ 

Done.

tasty-query/shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
```diff
@@ -384,6 +384,7 @@ private[pickles] class PickleReader {
         val pre = readPrefix()
         val designator = readMaybeExternalSymbolRef()
         designator match
+          case sym: PackageSymbol          => sym.packageRef
```
bishabosha, line 387: does this fix the test case about erasure of `java.lang.reflect.Field`?

bishabosha: oh no wait that was in tasty signatures, but I guess this should cover scala 2 also

bishabosha: I can add regression test for that in a new PR

tasty-query/shared/src/main/scala/tastyquery/Types.scala
```diff
@@ -642,8 +645,8 @@ object Types {
           case declaring: DeclaringSymbol =>
             val candidate = declaring.getDecl(name)
             candidate.getOrElse {
-              val msg = this match
-                case TermRef(_, name: SignedName) if declaring.memberPossiblyOverloaded(name) =>
+              val msg = this.name match
```
bishabosha, line 648: its interesting to note here that the `name` when designator was a `LookupIn` will always be a `SignedName`, because it will be cached on this access here, before designator is replaced by a symbol

bishabosha (approved):

PR #146, opened 2022-10-05
Clean up the public API, part 4

sjrd:
I believe this is the last of it. ;)

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Types.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Types.scala
@@ -444,7 +444,7 @@ object Types {
   /** Type proxies.
     * Each implementation is expected to redefine the `underlying` method.
     */
-  abstract class TypeProxy extends Type {
+  sealed abstract class TypeProxy extends Type {
 
     /** The type to which this proxy forwards operations. */
     def underlying(using Context): Type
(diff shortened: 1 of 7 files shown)


tasty-query/shared/src/main/scala/tastyquery/Types.scala
```diff
@@ -477,26 +477,29 @@ object Types {
   }
 
   /** Non-proxy types */
-  abstract class GroundType extends Type {}
+  sealed abstract class GroundType extends Type
+
+  /** Superclass for temporary ground types used by custom algorithms. */
+  abstract class TempGroundType extends GroundType
```
bishabosha, line 483: better comment and name

tasty-query/shared/src/test/scala/tastyquery/ReadTreeSuite.scala
```diff
@@ -51,13 +51,47 @@ class ReadTreeSuite extends RestrictedUnpicklingSuite {
   private object TermRefInternal:
     def unapply(tpe: TermRef): Some[(Type, AnyDesignator)] = Some((tpe.prefix, tpe.designatorInternal))
 
+  /** Extractors for `Type`s. */
+  private object ty:
```
julienrf, line 55: What does “ty” stand for?

julienrf (approved):
I am not sure about the change at the top of `PickleReader.scala`, but otherwise it looks good to me.

tasty-query/shared/src/test/scala/tastyquery/ReadTreeSuite.scala
sjrd: "type", but short, because we need it in many places in this file, and making it longer would clutter the actual tests.

tasty-query/shared/src/test/scala/tastyquery/ReadTreeSuite.scala
```diff
@@ -51,13 +51,51 @@ class ReadTreeSuite extends RestrictedUnpicklingSuite {
...
+    *
+    * "`ty`" stands for "type", but we make it very short not to clutter the
+    * actual tests below.
+    */
+  private object ty:
+    object PackageRef:
+      def unapply(tpe: Types.PackageRef): Some[FullyQualifiedName] = Some(tpe.fullyQualifiedName)
+
+    object AppliedType:
+      def unapply(tpe: Types.AppliedType): Some[(Type, List[Type])] = Some((tpe.tycon, tpe.args))
```
bishabosha, line 64: these extractors can directly return the tuple

bishabosha (approved):

PR #148, opened 2022-10-05
Set up sbt-ci-release.

sjrd:
Following the instructions of https://github.com/sbt/sbt-ci-release
I have already set up the secrets on the repo.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,25 @@
+name: Release
+on:
+  push:
+    branches: [main]
+    tags: ["*"]
+jobs:
+  publish:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3.0.2
+        with:
+          fetch-depth: 0
+      - name: Set up JDK 8
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'adopt'
+          java-version: 8
+          cache: 'sbt'
+      - name: sbt ci-release
+        run: sbt ci-release
+        env:
+          PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
+name: Release
+on:
+  push:
+    branches: [main]
+    tags: ["*"]
```
julienrf, line 5: This will publish snapshots on every push to `main`, is that what you want?

.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+        with:
+          fetch-depth: 0
+      - name: Set up JDK 8
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'adopt'
+          java-version: 8
+          cache: 'sbt'
+      - name: npm install
+        run: npm install
```
julienrf, line 20: Why is this needed?

julienrf (commented):
Eventually, it would be great to also set `versionScheme` and set up bincompat checks (maybe use sbt-version-policy?)

.github/workflows/release.yml
sjrd: Yes. We should have early feedback on whether the publishing doesn't publish "too much", at the very least.

.github/workflows/release.yml
sjrd: We also use in the `ci.yml`, so I put it here for consistency of the environment.

But in theory it shouldn't be needed. Do you prefer if I remove it?

julienrf: I think it is better to omit it in this case (to avoid unnecessary computations)

sjrd: OK, I removed it.

julienrf (approved):

PR #152, opened 2022-10-06
Fix #126: Use findMember to resolve Name-based NamedType's.

### LINKED ISSUE #126: Could not find symbol for inherited type member
Here is the failing test:

```scala
package simple_trees

abstract class Parent {
  type TypeMember = String
}

class InheritedTypeMember extends Parent {
  def m: TypeMember = ""
}
```

```scala
  testWithContext("inherited type member") {
    val InheritedTypeMember = resolve(name"simple_trees" / tname"InheritedTypeMember").asClass
    val m = InheritedTypeMember.getDecl(name"m").get
    println(m.signature)
  }
```

```
Could not find symbol for name TypeMember/T: not a member of symbol[class InheritedTypeMember]
```

sjrd:


tasty-query/shared/src/test/scala/tastyquery/BaseUnpicklingSuite.scala
```diff
@@ -8,4 +8,6 @@ import tastyquery.testutil.TestPlatform
 
 abstract class BaseUnpicklingSuite extends munit.FunSuite {
   lazy val testClasspath: Future[Classpath] = TestPlatform.loadClasspath()
+
+  override def munitTimeout = scala.concurrent.duration.Duration.Inf
```
bishabosha, line 12: overlooked

sjrd: That was on purpose, in a dedicated commit, so that it remains there for all our future debugging sessions.

bishabosha (approved):

PR #162, opened 2022-10-13
Simplify management of declarations; remove *Internal variants.

sjrd:


tasty-query/shared/src/main/scala/tastyquery/reader/Loaders.scala
```diff
@@ -19,7 +19,7 @@ private[tastyquery] object Loaders {
   class MissingTopLevelTasty(root: Loader.Root) extends Exception(s"Missing TASTy for ${root.fullName}")
 
   object Loader:
-    private[tastyquery] case class Root private[Loaders] (pkg: PackageSymbol, rootName: SimpleName):
+    private[Loaders] case class Root private[Loaders] (pkg: PackageSymbol, rootName: SimpleName):
       def packages: IterableOnce[PackageSymbol] =
```
bishabosha, line 23: this method is now dead code - it used to be to validate that a tasty file's declared packages match the file name, but now it seems that check was dropped?

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -536,9 +470,49 @@ object Symbols {
...
+    final def getDecl(name: Name)(using Context): Option[Symbol] =
+      name match
+        case overloaded: SignedName =>
+          distinguishOverloaded(overloaded)
+        case name =>
+          myDeclarations.get(name) match
+            case Some(decls) =>
+              if decls.sizeIs == 1 then Some(decls.head)
+              else if decls.sizeIs > 1 then
+                throw MemberNotFoundException(this, name, s"unexpected overloads in $this: ${decls.mkString(", ")}")
```
bishabosha, line 499: This is not compatible with a class that has overloads without parameters:
e.g.
```scala
class Foo:
  def foo = 23
  def foo(i: Int) = i
```

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -536,9 +470,49 @@ object Symbols {
...
+      myDeclarations.get(overloaded.underlying) match
+        case Some(overloads) =>
+          overloads.find {
+            case decl: TermSymbol => decl.signature.exists(_ == overloaded.sig)
+            case _                => false
+          }
+        case None => None
+
+    final def declarations(using Context): List[Symbol] =
+      myDeclarations.values.toList.flatten
```
bishabosha, line 513: seeing `ensureDeclsInitialized` dropped (even tho it was a no-op) made me think that we don't have a check to ensure these methods are not called during initialisation

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -655,12 +618,37 @@ object Symbols {
...
+      getDecl(name).toList
+
+    private final def ensureRootsInitialized()(using Context): Unit =
+      if !rootsInitialized then
+        ctx.classloader.scanPackage(this)
+        rootsInitialized = true
+
+    final def getDecl(name: Name)(using Context): Option[Symbol] =
+      myDeclarations.get(name).orElse {
+        ensureRootsInitialized()
```
bishabosha, line 642: I would suggest making `ensureRootsInitialized` return boolean, then only include the fallback when the package was just scanned now

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -182,14 +182,6 @@ object Symbols {
       case scope: DeclaringSymbol => scope.hasOverloads(name)
       case _                      => false
 
-    private[tastyquery] final def memberPossiblyOverloaded(name: SignedName): Boolean = this match
```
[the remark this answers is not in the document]
bishabosha: but still, if we want to add  the debug message in again then we can figure out what support is needed - probably `getDecls` is enough

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
sjrd: That logic was already there before:
https://github.com/scalacenter/tasty-query/pull/162/files#diff-5977ded6850927706fcf3030a7eccf82551476d77a8bb2e7d962cae7b91bb81dL401-L405

We should definitely fix it, but I don't think it belongs in this PR.

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
sjrd: That wouldn't be correct. If you first do `getDecl("foo")`, roots will be initialized. But then you do `getDecl("bar")`, and then `ensureRootsInitialized()` will do nothing, but you *still* need to try `loadRoot(this, "bar")`.

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
sjrd: I can try to come up with something here if you want. We could have a flag that is set when doing `checkCompleted()`. Before it's set, we cannot call `getDecl`/`declarations`. After it's set, we cannot call `addDecl`. Would that make sense?

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
bishabosha: yes right

bishabosha (approved):
nice stuff

PR #163, opened 2022-10-16
First implementation of subtyping.

sjrd:


tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -147,10 +149,13 @@ object Symbols {
       Iterator.iterate(enclosingDecl)(_.enclosingDecl).takeWhile(s => s.maybeOuter.exists)
 
     private[tastyquery] final def isStatic: Boolean =
-      Iterator
-        .iterate(this)(_.owner.nn)
-        .takeWhile(s => s.maybeOuter.exists)
-        .forall(s => s.isPackage || s.isClass && s.name.toTypeName.wrapsObjectName)
+      owner == null || owner.isStaticOwner
```
bishabosha, line 152: NoSymbol also has null owner, but is not static

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -147,10 +149,13 @@ object Symbols {
...
+
+    @tailrec
+    private def isStaticOwner: Boolean = this match
+      case _: PackageSymbol => true
+      case cls: ClassSymbol => cls.name.wrapsObjectName && cls.owner.isStaticOwner
```
bishabosha, line 157: we can probably check the module flag instead

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -518,15 +559,56 @@ object Symbols {
       myTypeParams != null
 
     final def derivesFrom(base: Symbol)(using Context): Boolean =
-      base.isClass &&
-        ((this eq base)
-        //|| (baseClassSet contains base)
-        )
+      base.isClass && ((this == base) || linearization.contains(base))
```
bishabosha, line 562: this seems identical to isSubclass

tasty-query/shared/src/test/scala/tastyquery/SubtypingSuite.scala
```diff
@@ -0,0 +1,237 @@
...
+    assertNeitherSubtype(MatchErrorClass.typeRef, ProductClass.typeRef)
+  }
+
+  testWithContext("monomorphic-class-type-alias") {
+    assertEquiv(defn.ObjectType, defn.AnyRefType)
+
+    assertEquiv(defn.StringType, PredefPrefix.select(tname"String"))
+    assertNeitherSubtype(defn.StringType, ScalaPackageObjectPrefix.select(tname"Cloneable"))
+
+    assertEquiv(defn.StringType, PredefPrefix.select(tname"String"))
```
bishabosha, line 97: duplicated lines

sjrd:
I think this is ready for a first version.

@smarter may be interested to have a look as well. :p

sjrd:
@bishabosha I added a commit that showcases compiler-driven consistency checks for the subtyping tests. If it looks like a reaonsable "API" for that, I will fill in the rest of the test cases.

bishabosha:
> @bishabosha I added a commit that showcases compiler-driven consistency checks for the subtyping tests. If it looks like a reaonsable "API" for that, I will fill in the rest of the test cases.

This looks very cool!

sjrd:
I now expanded the consistency checks to all the tests for which it makes sense.

bishabosha (approved):
the test suite is very cool

tasty-query/shared/src/main/scala/tastyquery/Subtyping.scala
```diff
@@ -0,0 +1,306 @@
...
+      isSubtype(tp1, tp2.first) && isSubtype(tp1, tp2.second)
+
+    case tp2: ConstantType =>
+      tp1 match
+        case tp1: ConstantType =>
+          tp1.value == tp2.value
+        case _ =>
+          level2(tp1, tp2)
+
+    case _ =>
```
smarter, line 63: I don't see an equivalent of https://github.com/lampepfl/dotty/blob/ef962cc2ee8d7cb2e8aebb5391406c74465e29c1/compiler/src/dotty/tools/dotc/core/TypeComparer.scala#L344-L345

tasty-query/shared/src/main/scala/tastyquery/Subtyping.scala
```diff
@@ -0,0 +1,306 @@
...
+      level3(tp1, tp2)
+  end level2
+
+  private def level3(tp1: Type, tp2: Type)(using Context): Boolean = tp2 match
+    case tp2: TypeRef =>
+      tp2.symbol match
+        case cls2: ClassSymbol =>
+          if cls2.typeParams.isEmpty then
+            if isBottom(tp1) then true
+            else if tp1.isLambdaSub then false // should be tp1.hasHigherKind, but the scalalib does not like that
```
smarter, line 128: Does this comment apply to tasty-query too? :) (In fact, I wonder if the original comment is still true, there were some fixes in the standard library)

tasty-query/shared/src/main/scala/tastyquery/Subtyping.scala
```diff
@@ -0,0 +1,306 @@
...
+            case sym2: TypeSymbolWithBounds =>
+              // TODO Handle polymorphic type aliases (compareLower in TypeComparer)
+              level4(tp1, tp2)
+        end if
+
+      case _ =>
+        false
+  end compareAppliedType2
+
+  private def isSubArgs(args1: List[Type], args2: List[Type], tp1: AppliedType, tparams2: List[TypeParamInfo])(
```
smarter, line 197: tp1 is unused here (in the original implementation it looks like it's only used for capture calculus stuff)

PR #165, opened 2022-10-17
Stricter compiler options.

sjrd:
In particular, `-deprecation` and `-Xfatal-warnings`.


bishabosha (approved):
maybe version 3.2.0 also?

PR #166, opened 2022-10-20
Remove `ClassInfo`.

sjrd:
All the code paths involving `ClassInfo` were dead code. Now that we have a significant portion of subtyping, I am confident that its need will not resurface later.

This introduces an `AssertionError` path in `TypeRef.underlying`, which is not nice, but it also removes some `???` code paths in other places.

It would be worth investigating whether we could make `TypeRef` not a `TypeProxy`, but a cursory attempt met other issues.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Substituters.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Substituters.scala
@@ -39,7 +39,7 @@ private[tastyquery] object Substituters:
     end apply
   end SubstBindingMap
 
-  private final class SubstParamsMap(from: Binders, to: List[Type])(using Context) extends DeepTypeMap:
+  private final class SubstParamsMap(from: Binders, to: List[Type])(using Context) extends NormalizingTypeMap:
     def apply(tp: Type): Type =
       tp match
         case tp: ParamRef =>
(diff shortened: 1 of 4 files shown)


bishabosha:
> It would be worth investigating whether we could make TypeRef not a TypeProxy, but a cursory attempt met other issues.

wouldn't this be because in dotty's model a TypeRef has to refer to some underlying type, and for a class, that is its classinfo - and I guess bounds for an abstract type member

bishabosha (approved):
I don't like the assertion error in underlying, perhaps it could be identity? but then its mostly used for recursive algorithms so not good to silently loop

sjrd:
> > It would be worth investigating whether we could make TypeRef not a TypeProxy, but a cursory attempt met other issues.
> 
> wouldn't this be because in dotty's model a TypeRef has to refer to some underlying type, and for a class, that is its classinfo - and I guess bounds for an abstract type member

Yes, that's the idea in dotc; but it's unclear when it stops at the `TypeRef` and when it allows itself to follow to the `ClassInfo`. In our model, the "ground" representation of a class type *is* the `TypeRef`. All our algorithms dispatch on `TypeRef`s and then on the kind of `symbol` they have. The only places where use `TypeProxy`-related methods on a `TypeRef` rule out the `ClassSymbol` case beforehand.

bishabosha:
should we have a ClassRef?

sjrd:
Yeah, precisely, we cannot know at the time of building the `TypeRef` if it would become a `ClassRef` after resolution.

PR #170, opened 2022-10-20
Implement overriding relationships.

sjrd:


tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -208,7 +210,74 @@ object Symbols {
...
+    /** If false, this symbol cannot possibly participate in an override, either as overrider or overridee. */
+    final def canMatchInheritedSymbols(using Context): Boolean =
+      owner.isClass && memberCanMatchInheritedSymbols
+
+    /** If false, this class member cannot possibly participate in an override, either as overrider or overridee. */
+    private final def memberCanMatchInheritedSymbols(using Context): Boolean =
+      !isConstructor && !is(Private)
+
+    private final def isConstructor(using Context): Boolean =
+      name == nme.Constructor
```
bishabosha, line 248:
```suggestion
      is(Method) && name == nme.Constructor
```

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -208,7 +210,74 @@ object Symbols {
...
+
+    /** The symbol, in class `inClass`, that is overridden by this symbol, if any. */
+    final def overriddenSymbol(inClass: ClassSymbol)(using Context): Option[Symbol] =
+      if canMatchInheritedSymbols then
+        val ownerClass = owner.asClass
+        if ownerClass.isSubclass(inClass) then matchingDecl(inClass, siteClass = ownerClass)
+        else None
+      else None
```
bishabosha, line 256: This seems to match dotty more - so I guess it will match constructors when inClass is the owner
```suggestion
      else if owner.asClass == inClass then matchingDecl(inClass, siteClass = inClass)
      else None
```

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -208,7 +210,74 @@ object Symbols {
...
+            val site = siteClass.thisType
+            val targetType = self.declaredTypeAsSeenFrom(site)
+            candidates.find { candidate =>
+              // TODO Also check targetName here
+              candidate.declaredTypeAsSeenFrom(site).matchesLoosely(targetType)
+            }
+    end matchingDecl
+
+    /** If false, this symbol cannot possibly participate in an override, either as overrider or overridee. */
+    final def canMatchInheritedSymbols(using Context): Boolean =
```
bishabosha, line 240: does this need to be public API?
```suggestion
    private[tastyquery] final def canMatchInheritedSymbols(using Context): Boolean =
```

tasty-query/shared/src/main/scala/tastyquery/TypeOps.scala
```diff
@@ -72,4 +72,65 @@ private[tastyquery] object TypeOps:
       super.expandBounds(tp)
     }
   }
+
+  // Tests around `matches`
+
+  /** The implementation for `tp1.matches(tp2)`. */
+  final def matchesType(tp1: Type, tp2: Type, relaxed: Boolean)(using Context): Boolean = tp1.widen match
```
bishabosha, line 79: `relaxed` param is currently dead code - but this is a public API soo its not?

tasty-query/shared/src/main/scala/tastyquery/Types.scala
```diff
@@ -429,6 +429,37 @@ object Types {
...
+      * - Or neither type takes term or type parameters.
+      *
+      * (*) when matching with a Java method, we also regard Any and Object as equivalent parameter types. (TODO)
+      *
+      * This function will always use unsafe-nulls semamtics to check the types.
+      * This is because we are using a relaxed rule (ignoring `Null` types)
+      * to check overriding Java methods.
+      */
+    final def matches(that: Type)(using Context): Boolean =
+      TypeOps.matchesType(this, that, relaxed = false)
```
bishabosha, line 452: in dotty `relaxed = false` implies post erasure

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
sjrd: I don't think it's possible for a non-method to have `name == nme.Constructor` (which is `<init>`), is it?

tasty-query/shared/src/main/scala/tastyquery/TypeOps.scala
sjrd: `TypeOps` itself is `private[tastyquery]`, so this is not public.

I removed the `relaxed` parameter.

tasty-query/shared/src/main/scala/tastyquery/Types.scala
sjrd: It's `relaxed = !ctx.phase.erasedTypes` in dotc. So `true` after erasure, but `false` before.

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
sjrd: I made it `private`.

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
sjrd: I changed it and also added tests for that.

tasty-query/shared/src/main/scala/tastyquery/Types.scala
bishabosha: `ctx.phase.erasedTypes == false` before erasure, so `relaxed` should be `true`

sjrd: No OK I got it backwards. I've changed it back to the `relaxed = true` behavior.

bishabosha (approved):

PR #174, opened 2022-10-26
support java 45.3, multiple throws on method

bishabosha:
I ran a test case that forced all root symbols in the `jre/lib` jars - this exposed that multiple throws clauses are not supported, and also some classes shipped in rt.jar are compiled by java 1.1 (i.e. have classfile version 45.3)

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/reader/classfiles/ClassfileReader.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/reader/classfiles/ClassfileReader.scala
@@ -19,25 +19,33 @@ private[classfiles] final class ClassfileReader private () {
 
   transparent inline def pool(using pool: ConstantPool): pool.type = pool
 
-  def acceptHeader()(using DataStream): Unit = {
-    acceptMagicNumber()
-    acceptVersion()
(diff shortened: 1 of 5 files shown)


tasty-query/shared/src/main/scala/tastyquery/reader/classfiles/ClassfileReader.scala
```diff
@@ -19,25 +19,33 @@ private[classfiles] final class ClassfileReader private () {
 
   transparent inline def pool(using pool: ConstantPool): pool.type = pool
 
-  def acceptHeader()(using DataStream): Unit = {
-    acceptMagicNumber()
-    acceptVersion()
+  def acceptHeader(classOwner: DeclaringSymbol, classRoot: ClassData)(using DataStream): Unit = {
```
bishabosha, line 22: 0.3 release?

bishabosha: or I can remove these as they were only for improving the error debug message

sjrd: This class is `private[classfiles]`, so it's not part of the public API, so no need for 0.3. ;)

sjrd (approved):

PR #185, opened 2022-11-03
Better APIs to find known top-level and static things.

sjrd:


sjrd:
If that looks good, we can do follow up changes to replace `resolve` in the other test suites. I started with two small ones that show representative improvements.

bishabosha (approved):
nice

PR #186, opened 2022-11-03
Fix/companion lookup

### LINKED ISSUE #184: enable companion lookup methods for non static classes
currently you can only lookup the companion object of a class if the class owner is a package

guihardbastien:
fixes #184

### DIFF
--- a/tasty-query/shared/src/test/scala/tastyquery/TypeSuite.scala
+++ b/tasty-query/shared/src/test/scala/tastyquery/TypeSuite.scala
@@ -1256,4 +1256,25 @@ class TypeSuite extends UnrestrictedUnpicklingSuite {
     assert(clue(childToString.allOverriddenSymbols.toList) == List(superToString, objectToString))
     assert(clue(childToString.nextOverriddenSymbol) == Some(superToString))
   }
+
+  def companionClassFullCycle(path: DeclarationPath)(using Context, munit.Location): Unit = {
+    val cls: ClassSymbol = resolve(path).asClass
+    val moduleClass: ClassSymbol = resolve(path.asObj).asClass
+
+    assert(cls == moduleClass.companionClass.get)
+    assert(moduleClass.companionClass.get == cls)
+  }
+
+  testWithContext("companion-tests-module-value") {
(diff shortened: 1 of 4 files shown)


tasty-query/shared/src/test/scala/tastyquery/TypeSuite.scala
```diff
@@ -1256,4 +1256,26 @@ class TypeSuite extends UnrestrictedUnpicklingSuite {
...
+  def companionClassFullCycle(path: DeclarationPath)(using Context, munit.Location): Unit = {
+    val Class: ClassSymbol = resolve(path).asClass
+    val Object: ClassSymbol = resolve(path.asObj).asClass
+
+    assert(Object == Class.companionClass.get)
+    assert(Object.companionClass.get == Class)
+  }
+
+  testWithContext("companion-tests-module-value") {
+    val x: Any = name"companions" / tname"CompanionObject"
```
sjrd, line 1269: This is unused.

tasty-query/shared/src/test/scala/tastyquery/TypeSuite.scala
```diff
@@ -1256,4 +1256,26 @@ class TypeSuite extends UnrestrictedUnpicklingSuite {
     assert(clue(childToString.allOverriddenSymbols.toList) == List(superToString, objectToString))
     assert(clue(childToString.nextOverriddenSymbol) == Some(superToString))
   }
+
+  def companionClassFullCycle(path: DeclarationPath)(using Context, munit.Location): Unit = {
+    val Class: ClassSymbol = resolve(path).asClass
+    val Object: ClassSymbol = resolve(path.asObj).asClass
```
sjrd, line 1262: `Class` and `Object` are well-known things in `java.lang`, so those names are a bit confusing. Consider using `cls` and `moduleClass` instead, which would follow the convention used in the codebase.

sjrd (changes requested):
Very good tests, thanks.

Could you also squash all the commits in a single one. Merge commits in particular should be avoided in a PR branch.

sjrd (approved):
Congrats 🎉

PR #188, opened 2022-11-10
Fix #187: getDecl considers no-signature as an identifying feature.

### LINKED ISSUE #187: Cannot resolve references to overload without signature
With the following added definitions to `OverloadedApply.scala`:
```scala
  def foo: String = "foo"

  def callF = foo
```
The following `TypeSuite` test fails:
```scala
  testWithContext("apply-overloaded-not-method") {
    val OverloadedApplyClass = ctx.findTopLevelClass("simple_trees.OverloadedApply")

    val callSym = OverloadedApplyClass.findDecl(termName("callF"))

    val Some(callTree @ _: DefDef) = callSym.tree: @unchecked

    var callCount = 0

    callTree.walkTree { tree =>
      tree match
        case fooRef @ Select(_, SimpleName("foo")) =>
          callCount += 1
          val fooSym = fooRef.tpe.asInstanceOf[TermRef].symbol
          assert(clue(fooSym.paramRefss).isEmpty)
        case _ => ()
    }

    assert(callCount == 1)
  }
```
That's because our `getDecl` refuses to select the one overload that *doesn't* require a `SignedName` when it's given an unsigned name.

This should clearly be addressed. But at the same time, I believe we should make `getDecl` *more strict* for methods that require a `SignedName`. Currently, if it finds one and it happens to be the only overload, it will pick it. But that is fragile, as something that resolves today may not resolve in the future when there is one more overload. So we should return `None` in that case.

Of course, for convenience, we should add a *separate* method that allows to pick a non-overloaded term decl only from its unsigned name. It should have an explicit name.

sjrd:
With some cleanups before we get there.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
@@ -225,7 +225,7 @@ object Symbols {
         case self: TypeSymbol =>
           inClass.getDecl(self.name).filterNot(_.is(Private))
         case self: TermSymbol =>
-          val candidates = inClass.getDecls(self.name).asInstanceOf[List[TermSymbol]].filterNot(_.is(Private))
+          val candidates = inClass.getAllOverloadedDecls(self.name).filterNot(_.is(Private))
           if candidates.isEmpty then None
           else
             val site = siteClass.thisType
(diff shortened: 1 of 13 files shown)


tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -668,6 +677,65 @@ object Symbols {
...
+
+    /** Convenience method to get a non-overloaded decl from its unsigned name.
+      *
+      * If there are multiple or no overload with the given unsigned name, this
+      * method returns `None`.
+      *
+      * @throws java.lang.IllegalArgumentException
+      *   if the provided `name` is a [[SignedName]]
+      */
+    final def getNonOverloadedDecl(name: TermName)(using Context): Option[TermSymbol] =
```
bishabosha, line 715: I guess this isn't statically SimpleName for convenience?

sjrd: There are other `TermName`s that are not `SignedName`s, for example `DefaultGetterName`s.

bishabosha (approved):

PR #194, opened 2022-11-18
Fix #193: Replace all uses of accessibleThisType by thisType.

### LINKED ISSUE #193: isSubtype fails for nested classes
Problematic input:
```
package subtyping.paths

object NestedClasses:
  class SuperClass

  class Outer(x: Int):
    class Inner(y: Int) extends SuperClass

  val outer = new Outer(5)
  val inner = new outer.Inner(6)
end NestedClasses
```

Test that demonstrates failure:
```
testWithContext("simple-paths-in-subclasses".only) {
  import subtyping.paths.NestedClasses.*

  val paths = "subtyping.paths"
  val inner = ctx.findStaticTerm(s"$paths.NestedClasses.inner")

  val st = inner.declaredType.isSubtype(defn.IntType)        // problem is here
  Console.println(st)
}
```

The marked line triggers an infinite loop in `Symbols.scala`.

sjrd:
It turns out that all the places where we used `accessibleThisType` were supposed to use the (internal) `thisType` instead. We replace them all, and remove the now-dead-code `accessibleThisType`.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
@@ -803,13 +803,6 @@ object Symbols {
       }
     end findMember
 
-    /** Get the self type of this class, as if viewed from an external package */
-    private[tastyquery] final def accessibleThisType: TypeRef =
-      owner match
-        case pre: PackageSymbol => TypeRef(pre.packageRef, this)
-        case pre: ClassSymbol   => TypeRef(pre.accessibleThisType, this)
-        case _                  => TypeRef(NoPrefix, this)
-
     private var myTypeRef: TypeRef | Null = null
 
     private[tastyquery] final def typeRef(using Context): TypeRef =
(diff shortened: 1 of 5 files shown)


tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -150,6 +150,15 @@ object Symbols {
       case cls: ClassSymbol => cls.is(Module) && cls.owner.isStaticOwner
       case _                => false
 
+    private[Symbols] final def staticOwnerPrefix(using Context): Type = this match
+      case pkg: PackageSymbol =>
+        pkg.packageRef
+      case cls: ClassSymbol if cls.is(Module) =>
+        cls.owner.staticOwnerPrefix.select(cls.moduleValue.get)
```
bishabosha, line 157: `moduleValue` still needs patching to enable non-package owners

sjrd: I pushed an additional commit that addresses that.

sjrd:
Ping @bishabosha ?

bishabosha (approved):

PR #196, opened 2022-11-21
Add a method FlagSet.show.

sjrd:
For debugging purposes.


tasty-query/shared/src/main/scala/tastyquery/Flags.scala
```diff
@@ -16,59 +16,69 @@ object Flags:
 
     def |(otherFlags: FlagSet): FlagSet = bits | otherFlags.bits
     def &(otherFlags: FlagSet): FlagSet = bits & otherFlags.bits
+
+    def show: String =
+      val result = new java.lang.StringBuilder
+      for (flag, name) <- allFlags if is(flag) do result.append(" " + name)
```
bishabosha, line 22:
```suggestion
      for (flag, name) <- allFlags if is(flag) do result.append(" | " + name)
```

bishabosha (approved):
nice

PR #197, opened 2022-11-21
Fix #195: Handle Private and ParamAccessor in `findMember`.

### LINKED ISSUE #195: findMember does not handle private members correctly
Test source:
```scala
package inheritance

object PrivateOverrides:
  class Parent:
    val y = 2

  class Child(y: Int) extends Parent

  def testSetup(): Int =
    val child = new Child(4)
    child.y
end PrivateOverrides
```
Test case:
```scala

  testWithContext("private overrides") {
    val ParentClass = ctx.findStaticClass("inheritance.PrivateOverrides.Parent")
    val ChildClass = ctx.findStaticClass("inheritance.PrivateOverrides.Child")

    val setupMethod = ctx.findTopLevelModuleClass("inheritance.PrivateOverrides").findNonOverloadedDecl(name"testSetup")
    val setupMethodDef = setupMethod.tree.get.asInstanceOf[DefDef]
    //val Left(valDefs) = setupMethodDef.paramLists.head: @unchecked
    //val List(x, y, z) = valDefs.map(valDef => TermRef(NoPrefix, valDef.symbol))
    val child = findLocalValDef(setupMethodDef.rhs.get, name"child")

    val selectNode = findTree(setupMethodDef.rhs.get) {
      case select: Select => select
    }
    val selectTpe = selectNode.tpe.asInstanceOf[TermRef]
    assert(selectTpe.name == termName("y"))
    val selectSym = selectTpe.symbol
    assert(clue(selectSym.owner) == ParentClass)
  }
```

sjrd:
When looking directly at a class, we should find its `Private` members, but not ones inherited from superclasses.

Even for the class itself, we should never find `ParamAccessor`s. If a `ParamAccessor` is actually meant, it will be a symbol-based `TermRef`, not a name-based one.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
@@ -787,19 +787,20 @@ object Symbols {
     end baseTypeOf
 
     private[tastyquery] final def findMember(pre: Type, name: Name)(using Context): Option[Symbol] =
-      def lookup(parents: List[Type]): Option[Symbol] = parents match
-        case p :: ps =>
-          p.classSymbol.flatMap { parentCls =>
-            // val inherited = parentd.membersNamedNoShadowingBasedOnFlags(name, required, excluded | Private)
-            // denots1.union(inherited.mapInherited(ownDenots, denots1, thisType))
-            parentCls.findMember(pre, name) // lookup in parents of parent
-          }.orElse(lookup(ps))
-        case nil => None
+      def lookup(lin: List[ClassSymbol]): Option[Symbol] = lin match
+        case parentCls :: linRest =>
+          parentCls.getDecl(name) match
+            case Some(sym) if !sym.is(Private) =>
+              Some(sym)
+            case _ =>
+              lookup(linRest)
+        case Nil =>
+          None
       end lookup
 
-      getDecl(name).orElse {
+      getDecl(name).filter(!_.is(ParamAccessor)).orElse {
         if name == nme.Constructor then None
-        else lookup(parents)
+        else lookup(linearization.tail)
       }
     end findMember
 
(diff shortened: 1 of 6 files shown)


test-sources/src/main/scala/inheritance/SameTasty.scala
```diff
@@ -6,13 +6,16 @@ object SameTasty:
     type FooType
     def foo: FooType
     def getFoo(): FooType = foo
+    //def maybePrivateMember: Int = 5
```
bishabosha, line 9:
```suggestion
```

test-sources/src/main/scala/inheritance/SameTasty.scala
```diff
@@ -6,13 +6,16 @@ object SameTasty:
...
 
   class Child extends Parent:
     type FooType = Int
     def foo: FooType = 23
+    private def maybePrivateMember: Int = 6
```
bishabosha, line 14: this isn't tested

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -787,19 +787,20 @@ object Symbols {
...
+            case Some(sym) if !sym.is(Private) =>
+              Some(sym)
+            case _ =>
+              lookup(linRest)
+        case Nil =>
+          None
       end lookup
 
-      getDecl(name).orElse {
+      getDecl(name).filter(!_.is(ParamAccessor)).orElse {
```
bishabosha, line 801: should ParamAccessors be added to decls?

sjrd: I think so. They are true members of the class. In an execution model, we have to create fields for them, for example.

bishabosha (approved):

PR #204, opened 2022-11-22
Add tree categories for term trees and pattern trees.

sjrd:
`TermTree`s are the only ones that have a generic `tpe`. Most term trees accept only `TermTree`s as arguments. For example, the `fun` and `args` or an `Apply` tree are all `TermTree`s.

Restricting `tpe` to `TermTree`s allows to get rid of many uses of `NoType`.

The pattern of `CaseDef` is limited to `PatternTree`s. Some "physical" TASTy tags are shared between similarly-looking term trees and pattern trees. The unpickler tells them apart. This way, when looking at patterns, the semantics are clearer.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Trees.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Trees.scala
@@ -37,25 +37,6 @@ object Trees {
   }
 
   sealed abstract class Tree(val span: Span) {
-    private var myType: Type | Null = null
-
-    /** Compute the actual term of this tree. */
-    protected def calculateType(using Context): Type
-
-    /** The term type of this tree.
-      *
-      * If this tree is not a term (e.g., it is a definition), its `tpe` is
-      * `NoType`.
-      */
-    final def tpe(using Context): Type = {
-      val local = myType
-      if local != null then local
-      else
-        val computed = calculateType
-        myType = computed
-        computed
-    }
-
     def withSpan(span: Span): Tree
 
     protected def subtrees: List[Tree] = this match {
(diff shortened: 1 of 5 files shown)


tasty-query/shared/src/main/scala/tastyquery/Trees.scala
```diff
@@ -475,54 +465,58 @@ object Trees {
...
+  final case class SeqLiteral(elems: List[TermTree], elemtpt: TypeTree)(span: Span) extends TermTree(span) {
     protected final def calculateType(using Context): Type =
       defn.SeqTypeOf(elemtpt.toType)
 
     override final def withSpan(span: Span): SeqLiteral = SeqLiteral(elems, elemtpt)(span)
   }
 
   /** while (cond) { body } */
-  final case class While(cond: Tree, body: Tree)(span: Span) extends Tree(span) {
+  final case class While(cond: TermTree, body: Tree)(span: Span) extends TermTree(span) {
```
bishabosha, line 488:
```suggestion
  final case class While(cond: TermTree, body: TermTree)(span: Span) extends TermTree(span) {
```

tasty-query/shared/src/main/scala/tastyquery/Trees.scala
```diff
@@ -542,7 +536,8 @@ object Trees {
     *
     * { bindings; expr }
     */
-  final case class Inlined(expr: Tree, caller: Option[TypeIdent], bindings: List[Tree])(span: Span) extends Tree(span) {
+  final case class Inlined(expr: TermTree, caller: Option[TypeIdent], bindings: List[Tree])(span: Span)
```
bishabosha, line 539: no idea if this would type check, but maybe it makes sense
```suggestion
  final case class Inlined(expr: TermTree, caller: Option[TypeIdent], bindings: List[DefTree])(span: Span)
```

tasty-query/shared/src/main/scala/tastyquery/Trees.scala
```diff
@@ -475,54 +465,58 @@ object Trees {
...
     */
-  final case class Unapply(fun: Tree, implicits: List[Tree], patterns: List[Tree])(span: Span) extends Tree(span) {
-    protected final def calculateType(using Context): Type = NoType
-
+  final case class Unapply(fun: TermTree, implicits: List[TermTree], patterns: List[PatternTree])(span: Span)
+      extends PatternTree(span) {
     override final def withSpan(span: Span): Unapply = Unapply(fun, implicits, patterns)(span)
   }
 
+  final case class ExprPattern(expr: TermTree)(span: Span) extends PatternTree(span):
```
bishabosha, line 473:
```suggestion
  final case class ExprPattern(expr: Ident | Literal)(span: Span) extends PatternTree(span):
```

bishabosha (approved):

PR #210, opened 2022-11-28
Correct owner for static java members.

bishabosha:
Set privateWithin for package private definitions. Set flags for classes.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
@@ -103,18 +103,17 @@ object Symbols {
     final def tree(using Context): Option[DefiningTreeType] =
       myTree
 
-    private[tastyquery] final def withFlags(flags: FlagSet): this.type =
+    private[tastyquery] final def withFlags(flags: FlagSet, privateWithin: Option[Symbol]): this.type =
       if isFlagsInitialized then throw IllegalStateException(s"reassignment of flags to $this")
       else
         isFlagsInitialized = true
         myFlags = flags
+        myPrivateWithin = privateWithin
         this
 
-    private[tastyquery] final def withPrivateWithin(privateWithin: Symbol): this.type =
-      if myPrivateWithin.isDefined then throw new IllegalStateException(s"reassignment of privateWithin to $this")
-      else
-        myPrivateWithin = Some(privateWithin)
-        this
+    final def privateWithin: Option[Symbol] =
+      if isFlagsInitialized then myPrivateWithin
+      else throw IllegalStateException(s"flags of $this have not been initialized")
 
     final def flags: FlagSet =
       if isFlagsInitialized then myFlags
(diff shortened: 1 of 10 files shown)


tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -135,6 +134,11 @@ object Symbols {
         assert(false, s"cannot access owner, ${this.name} is local or not declared within any scope")
     }
 
+    /** What is the closest enclosing package of this symbol? Returns this if this is a package */
+    private[tastyquery] final def closestPackage: PackageSymbol = this match
+      case pkg: PackageSymbol => pkg
+      case sym: Symbol        => sym.owner.nn.closestPackage // owner of regular symbols is never null
```
sjrd, line 140:
```suggestion
      case pkg: PackageSymbol => pkg
      case sym: TermOrTypeSymbol => sym.owner.closestPackage
```
`Symbol` is partitioned into `PackageSymbol` and `TermOrTypeSymbol`, and the latter's `owner` is known to be non-null. :)

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -135,6 +134,11 @@ object Symbols {
         assert(false, s"cannot access owner, ${this.name} is local or not declared within any scope")
     }
 
+    /** What is the closest enclosing package of this symbol? Returns this if this is a package */
```
sjrd, line 137:
```suggestion
    /** The closest enclosing package of this symbol.
      * Returns this if this is a package.
      */
```
Don't frame Scaladoc comments as questions.

tasty-query/shared/src/test/scala/tastyquery/TypeSuite.scala
```diff
@@ -376,10 +376,53 @@ class TypeSuite extends UnrestrictedUnpicklingSuite {
...
+
+  testWithContext("bag-of-java-definitions[static]") {
+    val BagOfJavaDefinitionsClassMod = ctx.findTopLevelModuleClass("javadefined.BagOfJavaDefinitions")
+    val javadefinedPackage = ctx.findPackage("javadefined")
+
+    def testDef(name: TermName)(op: munit.Location ?=> TermSymbol => Unit)(using munit.Location): Unit =
+      op(BagOfJavaDefinitionsClassMod.findNonOverloadedDecl(name))
+
+    testDef(name"STATIC_INT") { STATIC_INT =>
+      assert(STATIC_INT.declaredType.isRef(defn.IntClass))
```
sjrd, line 402: Add
```suggestion
      assert(!STATIC_INT.isAnyOf(Flags.Private | Flags.Protected))
      assert(STATIC_INT.privateWithin == None)
```
?

Same for `defaultInt` just below, as well as in the test `"bag-of-java-definitions"`.

sjrd (changes requested):

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
bishabosha: thanks

sjrd (approved):

PR #216, opened 2022-12-01
Support reading all sorts of refinement types.

### LINKED ISSUE #6: Support term refinements
Example:

```scala
val polyId: [T] => T => T = [T] => (x: T) => x
val id: (x: Any) => x.type = x => x
def bar: reflect.Selectable { val qux: Int } = new reflect.Selectable { val qux: Int = 23 }

class Poly[Id <: [T] => T => T](id: Id):
  def apply[T](x: T): T = id(x)
class Dep[Id <: (x: Any) => x.type](id: Id):
  def apply(x: Any): x.type = id(x)
class Structural[Refl <: reflect.Selectable { val qux: Int }](refl: Refl):
  def qux: Int = refl.qux
```

### LINKED ISSUE #16: Support RECthis Type Tag
Currently, ~~RECtype~~ RECthis tag is not supported by TreeUnpickler.

### LINKED ISSUE #20: Support POLYtype Type Tag

### LINKED ISSUE #21: Support METHODtype Type Tag

### LINKED ISSUE #213: Support for refinement types
Overall, refinement types have holes everywhere. From loading of term refinements (#6) to actually handling them in type substitution and subtyping.

sjrd:
Including their members which can contain RECthis, METHODtype, and POLYtype.

Fix #6, fix #16, fix #20, fix #21.

This is not enough for #213 yet, since they are not handled correctly in member selection nor subtyping yet.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/reader/tasties/TreeUnpickler.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/reader/tasties/TreeUnpickler.scala
@@ -1019,33 +1019,19 @@ private[tasties] class TreeUnpickler(
     case BYNAMEtype =>
       reader.readByte()
       ExprType(readType)
+    case POLYtype =>
+      readLambdaType(_ => PolyType, name => name.toTypeName, _.readTypeBounds)
+    case METHODtype =>
+      // TODO Record the `mods` somehow (given, implicit, erased)
+      readLambdaType(_ => MethodType, name => name, _.readType)
     case TYPELAMBDAtype =>
-      val lambdaAddr = reader.currentAddr
-      reader.readByte()
-      val end = reader.readEnd()
-      val resultUnpickler = fork // remember where the result type is
-      skipTree() // skip the result type
-      val paramTypeBoundsUnpickler = fork // remember where the params are
-      val paramNames = reader.until(end) {
-        skipTree() // skip the type bounds
-        readName.toTypeName
-      }
-      def readParamTypeBounds()(using LocalContext): List[TypeBounds] =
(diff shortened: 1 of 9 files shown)


tasty-query/shared/src/main/scala/tastyquery/reader/tasties/TreeUnpickler.scala
```diff
@@ -1095,6 +1084,48 @@ private[tasties] class TreeUnpickler(
...
+    var mods = EmptyFlagSet
+    while reader.currentAddr != end do // avoid boxing the mods
+      reader.readByte() match
+        case IMPLICIT => mods |= Implicit
+        case ERASED   => mods |= Erased
+        case GIVEN    => mods |= Given
+
+    // Read infos -- skip names
+    def readParamInfos()(using LocalContext): List[PInfo] =
+      paramInfosUnpickler.reader.until(end) {
```
bishabosha, line 1116: this looks like it would try to readInfo when the next byte is the mods

bishabosha: this example causes `unexpected type tag GIVEN` error:
```scala
val givenRefinement = ??? : { def foo(using Int): Int }
```

bishabosha (changes requested):
there is a bug in reading the parameter infos of lambdas

tasty-query/shared/src/main/scala/tastyquery/reader/tasties/TreeUnpickler.scala
sjrd: Oh nice catch! It's fixed now.

bishabosha (approved):

PR #217, opened 2022-12-01
Read annotations from TASTy.

sjrd:
This is the bare minimum to read the annotations.

There will be followups for
* Convenience accessors for annotation info (class symbol, constant arguments, etc.)
* Correct treatment of `targetName`

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
@@ -4,6 +4,7 @@ import scala.annotation.tailrec
 
 import scala.collection.mutable
 
+import tastyquery.Annotations.*
 import tastyquery.Contexts.*
 import tastyquery.Exceptions.*
 import tastyquery.Flags.*
@@ -75,6 +76,7 @@ object Symbols {
     private var myFlags: FlagSet = Flags.EmptyFlagSet
     private var myTree: Option[DefiningTreeType] = None
     private var myPrivateWithin: Option[Symbol] = None
+    private var myAnnotations: List[Annotation] = Nil
 
     /** Checks that this `Symbol` has been completely initialized.
       *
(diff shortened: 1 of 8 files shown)


tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -111,6 +113,14 @@ object Symbols {
         myPrivateWithin = privateWithin
         this
 
+    private[tastyquery] final def addAnnotations(annots: List[Annotation]): this.type =
+      myAnnotations = myAnnotations ::: annots
```
bishabosha, line 117: this seems to allow setting multiple times, I suggest making the field nullable and then checking for null in the accessor

tasty-query/shared/src/main/scala/tastyquery/reader/tasties/TreeUnpickler.scala
```diff
@@ -296,6 +297,30 @@ private[tasties] class TreeUnpickler(
       case tpe                => throw TastyFormatException(s"unexpected type for readWithin: $tpe")
   end readWithin
 
+  private def readAnnotationsInModifiers(sym: Symbol, end: Addr)(using LocalContext): Unit =
```
bishabosha, line 300: any reason not to read the annotations in `readModifiers`?

bishabosha (approved):

tasty-query/shared/src/main/scala/tastyquery/reader/tasties/TreeUnpickler.scala
sjrd: Yes, I tried that first. It didn't work because we read the modifiers while creating all the Symbols. To read annotations, we need to read Trees, and we can only do that after all the Symbols of the file have been created.

sjrd:
I added a commit that adds annotations to the checks performed by `checkCompleted()`. I left TODO's for reading annotations from Scala 2 and Java.

bishabosha (approved):

PR #221, opened 2022-12-08
Fix findMember for param accessors again.

sjrd:
The situation is much more complex than initially thought. Param accessors *can* be referred to by name, using `ThisType` prefixes. Moreover, non-private param accessors are truly the public vals, so they must be accessible from anywhere.

* We fix `widenIfUnstable`, so that `ThisType` are not widened (among others).
* We provide convenience accessors for the symbols of Idents and Selects.
* We add a new `flags.isAllOf(required, butNotAnyOf = excluded)` method.
* With all that, we fix `ClassSymbol.findMember` for the above issues.

/cc @shardulc


sjrd:
Rebased (without changes) to get the new `versionPolicyCheck`.

tasty-query/shared/src/main/scala/tastyquery/Trees.scala
```diff
@@ -267,6 +267,11 @@ object Trees {
   final case class Ident(name: TermName)(tpe: Type)(span: Span) extends TermTree(span):
     protected final def calculateType(using Context): Type = tpe
 
+    def symbolOption(using Context): Option[TermSymbol | PackageSymbol] = tpe match
```
bishabosha, line 270: consider unwrapping the option

bishabosha (approved):

PR #223, opened 2022-12-09
Define all the special methods on `Any` and `j.l.Object`.

sjrd:


bishabosha (changes requested):
missing some even more magic methods

bishabosha (approved):

PR #233, opened 2022-12-21
Add a method ClassSymbol.sealedChildren.

sjrd:


tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -926,6 +926,51 @@ object Symbols {
...
+      *
+      * If `this.is(Sealed)` is false, this method returns the empty set.
+      *
+      * Otherwise, each element of the set is either:
+      *
+      * - a non-module class that extends this class (including enum class cases), or
+      * - a module value whose module class extends this class, or
+      * - a enum val case for this enum class.
+      */
+    final def sealedChildren(using Context): Set[ClassSymbol | TermSymbol] =
```
bishabosha, line 942: I think this should be ordered, e.g. for generating documentation etc.

bishabosha (changes requested):
I think its important to preserve declaration order on the children

sjrd:
OK, I made it a `List` and added guarantees about the ordering in the documentation.

bishabosha (approved):

PR #235, opened 2022-12-23
Add methods that follow the semantics of dynamic method calls.

sjrd:
* `runtimeImplementingSymbol` for regular method calls.
* `nextSuperSymbol` for unqualified `super` method calls.

Qualified `super[T]` method calls are statically resolved, so they do not need any special method.

/cc @shardulc

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
@@ -244,7 +244,7 @@ object Symbols {
       * @param siteClass
       *   The base class from which member types are computed
       */
-    private final def matchingDecl(inClass: ClassSymbol, siteClass: ClassSymbol)(using Context): Option[Symbol] =
+    protected final def matchingDecl(inClass: ClassSymbol, siteClass: ClassSymbol)(using Context): Option[Symbol] =
       this match
         case self: TypeSymbol =>
           inClass.getDecl(self.name).filterNot(_.is(Private))
@@ -261,7 +261,7 @@ object Symbols {
     end matchingDecl
 
     /** If false, this symbol cannot possibly participate in an override, either as overrider or overridee. */
-    private final def canMatchInheritedSymbols(using Context): Boolean =
+    protected final def canMatchInheritedSymbols(using Context): Boolean =
       owner.isClass && memberCanMatchInheritedSymbols
 
     /** If false, this class member cannot possibly participate in an override, either as overrider or overridee. */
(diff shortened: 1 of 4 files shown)


tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -417,6 +417,78 @@ object Symbols {
...
+      else
+        if !this.owner.isClass then
+          throw IllegalArgumentException(s"Nonsensical run-time implementation lookup for local symbol $this")
+        this
+    end runtimeImplementingSymbol
+
+    /** Resolves a dynamically bound `super` call for an object of class `runtimeClass`.
+      *
+      * This method returns the next symbol for a `super` call of the form
+      * `enclosingClass.super.foo` for an object of class `runtimeClass`.
```
shardulc, line 450: `enclosingClass.super.foo` is confusing because in Scala, inside a class `A`, you cannot refer to a _parent_ class `B` as `B.super.foo`; you can only use that to refer to an _enclosing_ class in the sense that `A` is an inner class in `B`. But in tasty-query both of those things are represented the same way, as the type of the `This` inside the `Super`. And for this method in particular, we do require `enclosingClass` to be a parent of `runtimeClass` and definitely not an outer class.

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -417,6 +417,78 @@ object Symbols {
...
+      *   or this symbol is private;
+      *   or this symbol is a constructor;
+      *   or `runtimeClass.isSubclass(this.owner.asClass)` is false;
+      *   or if `runtimeClass.isSubclass(enclosingClass)` is false
+      * @throws tastyquery.Exceptions.MemberNotFoundException
+      *   if no concrete matching symbol could be found
+      */
+    final def nextSuperSymbol(runtimeClass: ClassSymbol, enclosingClass: ClassSymbol)(using Context): TermSymbol =
+      if !canMatchInheritedSymbols then
+        throw IllegalArgumentException(s"Invalid dynamic super call to statically bound symbol $this")
```
shardulc, line 467: How about indicating that the symbol is private or a constructor? I am not sure I would understand what "statically bound" means if I see this error message :‌D

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -417,6 +417,78 @@ object Symbols {
...
+
+    /** Common loop for `runtimeImplementingSymbol` and `nextSuperSymbol`. */
+    @tailrec
+    private def findRuntimeSymbolCommon(runtimeClass: ClassSymbol, linearization: List[ClassSymbol])(
+      using Context
+    ): TermSymbol =
+      linearization match
+        case inClass :: linearizationTail =>
+          matchingDecl(inClass, siteClass = runtimeClass) match
+            case Some(matchingSym) if !matchingSym.is(Abstract) =>
```
shardulc, line 485: I'm still a little skeptical of filtering out all abstract matching symbols here. But I cannot think of a case in which you need to find an abstract method when resolving a method call at runtime. With super calls inside traits, it doesn't happen because they are replaced with super _accessor_ calls, which although abstract, you never need to actually find, because you can find their referent just from the name. Maybe some edge case with abstract classes?

test-sources/src/main/scala/simple_trees/SuperDoubleDiamond.scala
```diff
@@ -0,0 +1,30 @@
...
+  // Lin = AB, B, A, AnyRef
+  trait AB extends A, B:
+    override def foo = super.foo
+
+  // Lin = AAABBB, BB, AB, B, AA, A, AnyRef
+  class AAABBB extends AA, AB, BB:
+    override def foo = super.foo
+
+  // Lin = AABBAB, AB, BB, B, AA, A, AnyRef
+  class AABBAB extends AA, BB, AB
```
shardulc, line 29: Just for exhaustiveness I would put in the third possible permutation too, `ABAABB`.

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
shardulc: I notice that the same `B.super.foo` notation is used in the docstring of `Super` too, perhaps it should be clarified there too. And maybe we should add a test for `nextSuperSymbol` that actually uses an inner class and demonstrates that the linearization of the inner class does not matter when resolving `Outer.super.foo` and vice versa.

tasty-query/shared/src/test/scala/tastyquery/ReadTreeSuite.scala
```diff
@@ -1083,6 +1083,55 @@ class ReadTreeSuite extends RestrictedUnpicklingSuite {
...
+                  Select(
+                    This(TypeIdent(TypeName(SimpleName("AB")))),
+                    PrefixedName(
+                      NameTags.SUPERACCESSOR,
+                      ExpandedName(
+                        NameTags.EXPANDED,
+                        ExpandedName(NameTags.EXPANDPREFIX, _, SimpleName("AB")),
+                        SimpleName("foo")
+                      )
+                    )
```
shardulc, line 1124: Do you also want to check that the other type of `PrefixedName` appears in the trees? The one corresponding to `super[Mixin].foo` of the form `PrefixedName(_, ExpandedName(EXPANDPREFIX, "foo", "Mixin"))`.

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
sjrd: All you say is true, but I don't see why it contradicts the use of `enclosingClass` here. Precisely, it must be enclosing the `super` call for it to be valid. Now tasty-query has no notion of "where" we "are", so we can give it an enclosing class and tasty-query has to trust that it is meaningful.

For the same reason, a test with an inner class would not change anything. We would pass exactly the same arguments to this method whether we "are" in the inner class or not. As far as a test goes, we would be copying the same thing twice without any difference.

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
sjrd: If we did *not* exclude abstract methods, what would you then do? You can't actually call an abstract method, since there is no body.

test-sources/src/main/scala/simple_trees/SuperDoubleDiamond.scala
sjrd: There are 3! = 6 permutations. Should I include them all?

shardulc: Well I was thinking that since `AA` and `BB` are positioned symmetrically in the inheritance graph, there are effectively only three permutations. But if you think the existing two permutations should already cover all situations then it’s fine.

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
shardulc: Fair… like I said, I still can’t think of a case in which it would be necessary, so this is probably correct as is. I mentioned abstract classes because that is something we didn’t talk about when we were discussing these methods.

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
shardulc: I entirely agree, I’m not saying it contradicts the use of `enclosingClass`. I’m saying it contradicts the use of `enclosingClass.super.foo`. Because if you have a class `B` with an inner class `A`, then `B` is a valid "enclosing class" of a `super` in the body of `A` which we could write as `B.super.foo`. But to evaluate `B.super.foo`, the runtime class must inherit from `B`, not from `A`, right? I’m just saying that writing it as `enclosingClass.super.foo` is confusing because it is not sufficient for `enclosingClass` to be an "enclosing class" of the `super`, but in addition it must be a superclass of the `runtimeClass`, which may not necessarily inherit from the class directly enclosing the `super` (`A` in our example).

sjrd:
After thinking more about this, I think it's better to expose (a safer variant of) `matchingDecl` in the public API, and let users of the API implement the two methods proposed here on their own. They are pretty specific to the interpreter use case, after all.

PR #249, opened 2023-02-08
Be even more tolerant when typing polymorphic `Apply(New, ...)`.

sjrd:
I was not able to reproduce the issue in a standalone test, but apparently it is possible to have *both* a `New` with an `AppliedType` (hence, monomorphic) wrapped in a `TypeApply`. This would attempt to reapply type parameters to an already monomorphic type, which crashed.

We are now a bit more tolerant, and apply type parameter lists to the type of the `New` as long as we have a polymorphic type. We stop applying further type parameter lists when the type is fully monomorphic.

---

The first commit contains more tests that were an attempt at reproducing the above issue. They don't, but the tests are valuable anyway.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Trees.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Trees.scala
@@ -323,15 +323,17 @@ object Trees {
       * of the result type of the constructor method (which is `Unit`).
       *
       * We would like to say that we should always use the `tpe` of the `New`
-      * node. However, that is not correct when the class is polymorphic *and*
-      * the type arguments are inferred! In that situation, the `New` node will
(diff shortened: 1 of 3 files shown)


bishabosha (approved):
it'd be interesting to document how this crashed originally?

sjrd:
> it'd be interesting to document how this crashed originally?

Things like this:
```
tastyquery.Exceptions$InvalidProgramStructureException: found type constructor
AppliedType(
  TypeRef(PackageRef(langoustine.lsp), symbol[class lsp.StringEnum]),
  List(TypeRef(ThisType(TypeRef(PackageRef(langoustine.lsp), enumerations)), MonikerKind))
)
without type params in AppliedType
```

PR #252, opened 2023-02-08
Better computation of top-level v nested classfiles in Loaders.

sjrd:
The previous computation was not able to deal with top-level classes containing `$` signs. Now, we only consider a class to be nested (or companion) if some other class in the same directory is a *prefix* of it.

When only `.tasty` files are present, all classes are top-level.


bishabosha (approved):
nice tests :)

PR #253, opened 2023-02-08
Handle SHAREDterm's in `readPattern`.

sjrd:
And with that, we solve the last issue caused by tasty-query in the langoustine project.
There will remain one issue to fix on tasty-mima's side.


bishabosha (approved):
looks good (I checked the tasty of `test-sources/src/main/scala/simple_trees/ForExpressions.scala`)

PR #256, opened 2023-02-10
Consolidate Trees for type bounds and type definitions.

sjrd:
We create a new category of trees, `TypeDefinitionTree`, for the rhs of type definitions: type members, type parameters and type captures. They isolate the weird, context-dependent meaning of TYPEBOUNDS, TYPEBOUNDStpt (sometimes wrapped in LAMBDAtpt) from the hierarchy of `Type`s and the corresponding `TypeTree`s.

The `TypeDefinitionTree`s are read from TASTy using a few isolated functions, and they are transformed depending on their context into `TypeBounds`, `WildcardTypeBounds`, `TypeMemberDefinition`s, etc.

These changes allow to completely remove `BoundedType` and `NamedTypeBounds` from the hierarchy of `Type`s, so that we are certain they do not leak to the semantic areas of the API.


bishabosha (approved):
nice simplification

PR #257, opened 2023-02-14
Do not wrap the result type of paramless methods in ExprType.

sjrd:
Reserve `ExprType` for by-name params only.

This is closer to the contents of TASTy and of pickles, and is also closer to a user understanding of the language.

We then rename `ExprType` to `ByNameType`.


bishabosha:
doesnt ExprType signal that the value is evaluated each time?

sjrd:
No, it's not remotely enough to know that. When you have `f(x)` an you compute `f.tpe.appliedTo(x.tpe)`, you never receive an `ExprType`, although `f(x)` is clearly evaluated each time.

bishabosha (approved):

PR #258, opened 2023-02-14
Make signature, targetName and signedName more user-friendly.

sjrd:
We make `signature` and `targetName` public. All terms now have a `signature` (it is non-optional). For fields, it is the signature of the JVM/IR getter method, which is also the descriptor of the underlying JVM field.

In contrast, `signedName` does not return a `SignedName` for parameterless methods anymore. It returns a `SignedName` if and only if a `SignedName` is required to find the symbol through `getDecl`. To access the JVM signature of a parameterless method, `signature` can be used.

---

~~Based on #257.~~

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
@@ -349,8 +349,9 @@ object Symbols {
     private var myDeclaredType: Type | Null = null
 
     // Cache fields
-    private var mySignature: Option[Signature] | Null = null
+    private var mySignature: Signature | Null = null
     private var myTargetName: TermName | Null = null
+    private var mySignedName: TermName | Null = null
     private var myParamRefss: List[Either[List[TermParamRef], List[TypeParamRef]]] | Null = null
 
     protected override def doCheckCompleted(): Unit =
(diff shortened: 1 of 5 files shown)


bishabosha (approved):

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -791,7 +801,7 @@ object Symbols {
         case Some(overloads) =>
           overloads.find {
             case decl: TermSymbol =>
-              decl.signature.exists(_ == overloaded.sig) && decl.targetName == overloaded.target
+              overloaded == decl.signedName
```
bishabosha, line 804: this seems like it would fail for any parameterless def
```scala
package bar

class Bar:
  def bar = 1
  def bar(x: Int) = x

  def test =
    val b = new Bar
    b.bar
    b.bar(1)
```
in the above example `b.bar` is encoded with a `SELECTin`, this will have a signed name, so will resolve its symbol by a call to `getDecl` with a signed name, so delegating to this method with a signed name, thus it will never match

bishabosha (dismissed):
I can't approve unless my example is tested as working (resolve the symbol of `b.bar`)

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
bishabosha: actually dismiss this, I misread the TASTy, actually this would be a plain SELECT

bishabosha (approved):

PR #259, opened 2023-02-15
Fix #213: Support refinement types.

### LINKED ISSUE #213: Support for refinement types

sjrd:
TODO: Decent commit message for the last commit.


tasty-query/shared/src/main/scala/tastyquery/Trees.scala
```diff
@@ -638,7 +654,27 @@ object Trees {
...
+              case _ =>
+                throw InvalidProgramStructureException(s"Unexpected rhs for type refinement '$name': $rhs")
+            end refinedBounds
+            TypeRefinement(parent, name, refinedBounds)
+
+          case ValDef(name, tpt, _, _) =>
+            TermRefinement(parent, name, tpt.toType)
+
+          case DefDef(name, paramClauses, resultType, _, _) =>
+            TermRefinement(parent, name, ParamsClause.makeDefDefType(paramClauses, resultType))
```
bishabosha, line 675: I believe that this does not distinguish `Foo { val x: Int }` from `Foo { def x: Int }`, in the future we may want to add an api here for consumers to tell the difference (e.g. for detecting "caching" opportunities)

bishabosha (approved):

PR #264, opened 2023-03-07
Fix #263: Always take .tasty files as top-level classes.

### LINKED ISSUE #263: Top-level package object is shadowed by class of same name as file name
```scala
// File name: ClassAndPackageObjectSameName.scala

package simple_trees

class ClassAndPackageObjectSameName

def topLevelMethod(): Int = 1
```
The `ClassAndPackageObjectSameName$package` package object and its object class cannot be found. They are shadowed by the class of the same name.

This regressed in 0.5.8, likely in 299aa90fd74fb147f28a97d7980279cdb1e18ae3.

sjrd:


adpi2 (approved):
Thanks for the fix

PR #266, opened 2023-03-16
Implement match types.

sjrd:


tasty-query/shared/src/main/scala/tastyquery/TypeMatching.scala
```diff
@@ -0,0 +1,310 @@
...
+  /** Does `tycon` have a field with type `tparam`? */
+  private def typeparamCorrespondsToField(tycon: Type, tparam: TypeParamInfo)(using Context): Boolean =
+    productSelectorTypes(tycon).exists {
+      case tp: TypeRef => tp.isLocalRef(tparam)
+      case _           => false
+    } /*|| tycon.derivesFrom(defn.PairClass)*/
+  end typeparamCorrespondsToField
+
+  private def productSelectorTypes(tycon: Type)(using Context): Iterator[Type] =
+    Iterator
```
bishabosha, line 301: questionable, why not look for the case Accessors?

Apparently this is what scala 3 does though

bishabosha (approved):
initial work looks good, but there should be tests added for match type cases over generic tuple types and enum singleton cases

tasty-query/shared/src/test/scala/tastyquery/SubtypingSuite.scala
```diff
@@ -830,4 +834,228 @@ class SubtypingSuite extends UnrestrictedUnpicklingSuite:
...
+
+    assertEquiv(matchTypeEnums.appliedTo(ctx.findStaticTerm(s"$MyEnumPrefix.Singleton2").staticRef), defn.StringType)
+      .withRef[MatchTypeEnums[MyEnum.Singleton2.type], String]
+
+    assertEquiv(matchTypeEnums.appliedTo(ParametricTycon.appliedTo(defn.StringType)), defn.StringType)
+      .withRef[MatchTypeEnums[MyEnum.Parametric[String]], String]
+  }
+
+  testWithContext("match-types-with-old-style-enums") {
+    import subtyping.TypesFromTASTy.MyEnum
```
bishabosha, line 1038:
```suggestion
    import subtyping.TypesFromTASTy.OldStyleEnum
```
plus update the other references

sjrd: Oops, indeed. Fixed.

bishabosha (approved):

PR #268, opened 2023-03-23
Implement subtyping for RefinedTypes.

sjrd:


tasty-query/shared/src/main/scala/tastyquery/Subtyping.scala
```diff
@@ -275,6 +279,27 @@ private[tastyquery] object Subtyping:
...
+        false
+
+      case ResolveMemberResult.TermMember(symbols, tpe) =>
+        tp2 match
+          case tp2: TermRefinement => isSubtype(tpe, tp2.refinedType)
+          case _: TypeRefinement   => throw AssertionError(s"found term member for $tp2 in $tp1")
+
+      case ResolveMemberResult.ClassMember(cls) =>
+        tp2 match
+          case tp2: TypeRefinement => tp2.refinedBounds.contains(tp1.select(cls))
```
bishabosha, line 294: does this still work for generic classes?

tasty-query/shared/src/main/scala/tastyquery/Subtyping.scala
```diff
@@ -312,6 +337,9 @@ private[tastyquery] object Subtyping:
           tp2.typeParams.lengthCompare(tp1.paramRefs) == 0
             && isSubtype(tp1.resultType, tp2.appliedTo(tp1.paramRefs))
 
+    case tp1: RefinedType =>
+      isSubtype(tp1.parent, tp2)
```
bishabosha, line 341: can tp1 have incompatible refinements with tp2? e.g. if they have the same class type, but tp1 introduces a override

bishabosha: revisiting, I guess no because refinements have to be subtype of their base class

sjrd: It doesn't matter. A refinement can only *improve* its parent. In other words, it is always true that `P { refinement } <: P`. Therefore, if `P <: tp2`, by transitivity of subtyping it is always correct to say that `P { refinement } <: tp2`.

(if it does have incompatible with *itself*, it effectively becomes `Nothing`, and so it is also trivially a subtype of `tp2`, so this is fine as well)

tasty-query/shared/src/main/scala/tastyquery/Subtyping.scala
bishabosha: Now I have no idea what to think:

```scala
class Foo:
  class Bar[T]

val works = summon[Foo { type Bar[T] <: String } <:< Foo]
```

bishabosha: at least it doesnt work the other way

bishabosha (approved):
I'd like to see a test for a refinement of a generic class type

PR #270, opened 2023-03-24
Add a method Type.lookupMember(name: Name): Option[NamedTyped].

sjrd:
That allows userland code to *query* a Type for a member by its name, without having to a priori construct a NamedType and catch `MemberNotFoundException`.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Types.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Types.scala
@@ -198,6 +198,24 @@ object Types {
     final def select(name: TypeName)(using Context): TypeRef =
       TypeRef(this, name)
 
+    final def lookupMember(name: Name)(using Context): Option[NamedType] =
+      resolveMember(name, this) match
+        case ResolveMemberResult.NotFound =>
+          None
+        case resolved: ResolveMemberResult.TermMember =>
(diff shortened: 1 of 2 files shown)


tasty-query/shared/src/main/scala/tastyquery/Types.scala
```diff
@@ -198,6 +198,24 @@ object Types {
...
+          None
+        case resolved: ResolveMemberResult.TermMember =>
+          Some(TermRef.fromResolved(this, resolved))
+        case resolved: ResolveMemberResult.ClassMember =>
+          Some(TypeRef.fromResolved(this, resolved))
+        case resolved: ResolveMemberResult.TypeMember =>
+          Some(TypeRef.fromResolved(this, name.asInstanceOf[TypeName], resolved))
+    end lookupMember
+
+    final def lookupMember(name: TermName)(using Context): Option[TermRef] =
```
bishabosha, line 213: one question: what should happen when you call this on a `PackageRef`, and its the name of a package?

bishabosha: Ok it produces `ResolveMemberResult.NotFound`

bishabosha (approved):

PR #275, opened 2023-03-29
Fix more issues discovered by the checker

sjrd:


tasty-query/shared/src/test/scala/tastyquery/TypeSuite.scala
```diff
@@ -1124,6 +1124,89 @@ class TypeSuite extends UnrestrictedUnpicklingSuite {
...
+      assertAnnotatedSeqOfInt(formal)
+      assertRepeatedOfInt(typed)
+      assertSeqOfInt(actual.widen)
+    }
+
+    locally {
+      val dd = getDefOf("givesSeqToJava")
+      val (formal, typed, actual) = extractFormalTypedActualParamTypes(dd.rhs.get)
+
+      assertArrayOfInt(formal)
```
bishabosha, line 1195: Does dotc also type it as a plain array? - if not then we should look for the `ACC_VARARGS` flag on the method

bishabosha: The answer is no :) ClassfileParser in dotty wraps `Array[Foo]` in Repeated if the method has `ACC_VARARGS` and it is the last parameter

bishabosha (approved):

sjrd:
I'll follow up with a fix for the Java definition-site varargs.

PR #277, opened 2023-03-30
Handle the ACC_VARARGS flag on Java methods.

sjrd:


tasty-query/shared/src/main/scala/tastyquery/reader/classfiles/ClassfileParser.scala
```diff
@@ -227,16 +228,36 @@ private[reader] object ClassfileParser {
...
+  private def patchForVarargs(sym: TermSymbol, tpe: Type)(using Context): Type =
+    tpe match
+      case tpe: MethodType if tpe.paramNames.sizeIs >= 1 =>
+        defn.internalRepeatedAnnotClass match
+          case Some(annotClass) =>
+            val patchedLast = AnnotatedType(tpe.paramTypes.last, TQAnnotation(annotClass))
+            tpe.derivedLambdaType(tpe.paramNames, tpe.paramTypes.init :+ patchedLast, tpe.resultType)
+          case None =>
+            // Warn here? How?
+            tpe
```
bishabosha, line 254: This seems like it should be an error, its incorrectly typed otherwise

sjrd: But if we do that, we cannot load Java/Scala 2-only classpaths anymore, which would be a pity.

sjrd: Well ... even if that's what it does, it doesn't match what it does at definition site for Scala 2 and 3. I think we should be more consistent.

bishabosha: ok

PR #278, opened 2023-03-30
Fix weirdly recursive base type computation.

sjrd:


tasty-query/shared/src/main/scala/tastyquery/TypeOps.scala
```diff
@@ -7,10 +7,14 @@ import tastyquery.TypeMaps.*
...
-    new AsSeenFromMap(pre, cls).apply(tp)
+    pre match
+      case NoPrefix                        => tp
+      case pre: ThisType if pre.cls == cls => tp // This is necessary to cut down infinite recursions
+      case pre: Type                       => new AsSeenFromMap(pre, cls).apply(tp)
+  end asSeenFrom
 
   /** The TypeMap handling the asSeenFrom */
-  class AsSeenFromMap(pre: Prefix, cls: Symbol)(using Context) extends ApproximatingTypeMap {
+  class AsSeenFromMap(pre: Type, cls: Symbol)(using Context) extends ApproximatingTypeMap {
```
bishabosha, line 17: perhaps this should be private - as it shouldn't be constructed with the wrong ThisType

bishabosha (approved):

PR #283, opened 2023-04-06
Implement capture conversion in subtyping.

sjrd:


smarter (approved):
Nice! Would it be useful to add a test exercising the class shadowing support?

sjrd:
Yes. I'll do that when I get back from vacation. In the meantime, I'll merge this for now.

PR #290, opened 2023-04-24
Read TERMREFin nodes.

sjrd:
I found those in the parboiled codebase, but I am not sure how to create a reproduction for them.


bishabosha:
here's a test case: https://github.com/scala/scala/blob/2.13.x/test/tasty/run/src-3/tastytest/TermRefIns.scala

sjrd:
Thanks. Added the test case.

bishabosha (approved):

PR #292, opened 2023-04-26
Implement erasure for the remaining TypeProxy's.

sjrd:
In particular, for `RecType`s.

This makes the scala-debug-adapter CI pass again.


adpi2 (approved):
Thanks

PR #300, opened 2023-05-26
The Great Types Refactoring: Adapt our Types hierarchy to the spec.

sjrd:
A `Type` is now what the spec calls a type. It does not include `WildcardTypeBounds`, `MethodicType` nor `PackageRef` anymore. For that, the super traits `TypeOrMethodic`, `TypeOrWildcard`, or `NonEmptyPrefix` must be used instead.

`TypeOrWildcard`s can only be used as type arguments of `AppliedType`s. `TypeOrMethodic`s are used as the underlying type of term members.

Arbitrary terms have `TermType`s. These are real `Type`s, `MethodicType`s or `PackageRef`s.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Definitions.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Definitions.scala
@@ -45,13 +45,13 @@ final class Definitions private[tastyquery] (ctx: Context, rootPackage: PackageS
   val ObjectType: TypeRef = TypeRef(javaLangPackage.packageRef, typeName("Object"))
 
   val ArrayTypeUnapplied: TypeRef = TypeRef(scalaPackage.packageRef, typeName("Array"))
-  def ArrayTypeOf(tpe: Type): AppliedType = AppliedType(ArrayTypeUnapplied, List(tpe))
+  def ArrayTypeOf(tpe: TypeOrWildcard): AppliedType = AppliedType(ArrayTypeUnapplied, List(tpe))
 
   val SeqTypeUnapplied: TypeRef = TypeRef(scalaCollectionImmutablePackage.packageRef, typeName("Seq"))
-  def SeqTypeOf(tpe: Type): AppliedType = AppliedType(SeqTypeUnapplied, List(tpe))
+  def SeqTypeOf(tpe: TypeOrWildcard): AppliedType = AppliedType(SeqTypeUnapplied, List(tpe))
 
   val RepeatedTypeUnapplied: TypeRef = TypeRef(scalaPackage.packageRef, tpnme.RepeatedParamClassMagic)
-  def RepeatedTypeOf(tpe: Type): AppliedType = AppliedType(RepeatedTypeUnapplied, List(tpe))
+  def RepeatedTypeOf(tpe: TypeOrWildcard): AppliedType = AppliedType(RepeatedTypeUnapplied, List(tpe))
 
   val IntType: TypeRef = TypeRef(scalaPackage.packageRef, typeName("Int"))
   val LongType: TypeRef = TypeRef(scalaPackage.packageRef, typeName("Long"))
@@ -66,7 +66,7 @@ final class Definitions private[tastyquery] (ctx: Context, rootPackage: PackageS
   val StringType: TypeRef = TypeRef(javaLangPackage.packageRef, typeName("String"))
 
   val UnappliedClassType: TypeRef = TypeRef(javaLangPackage.packageRef, typeName("Class"))
-  def ClassTypeOf(tpe: Type): AppliedType = AppliedType(UnappliedClassType, List(tpe))
+  def ClassTypeOf(tpe: TypeOrWildcard): AppliedType = AppliedType(UnappliedClassType, List(tpe))
 
   val ThrowableType: TypeRef = TypeRef(javaLangPackage.packageRef, typeName("Throwable"))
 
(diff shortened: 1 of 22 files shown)


bishabosha:
needs a rebase

sjrd:
Rebased.

bishabosha:
needs another rebase :)

tasty-query/shared/src/main/scala/tastyquery/Definitions.scala
```diff
@@ -45,13 +45,13 @@ final class Definitions private[tastyquery] (ctx: Context, rootPackage: PackageS
...
-  def ArrayTypeOf(tpe: Type): AppliedType = AppliedType(ArrayTypeUnapplied, List(tpe))
+  def ArrayTypeOf(tpe: TypeOrWildcard): AppliedType = AppliedType(ArrayTypeUnapplied, List(tpe))
 
   val SeqTypeUnapplied: TypeRef = TypeRef(scalaCollectionImmutablePackage.packageRef, typeName("Seq"))
-  def SeqTypeOf(tpe: Type): AppliedType = AppliedType(SeqTypeUnapplied, List(tpe))
+  def SeqTypeOf(tpe: TypeOrWildcard): AppliedType = AppliedType(SeqTypeUnapplied, List(tpe))
 
   val RepeatedTypeUnapplied: TypeRef = TypeRef(scalaPackage.packageRef, tpnme.RepeatedParamClassMagic)
-  def RepeatedTypeOf(tpe: Type): AppliedType = AppliedType(RepeatedTypeUnapplied, List(tpe))
+  def RepeatedTypeOf(tpe: TypeOrWildcard): AppliedType = AppliedType(RepeatedTypeUnapplied, List(tpe))
```
bishabosha, line 54: does this really have a wildcard sometimes?

tasty-query/shared/src/main/scala/tastyquery/TypeMaps.scala
```diff
@@ -301,12 +294,12 @@ private[tastyquery] object TypeMaps {
...
               case _                               => tryWiden(tp, preHi)
             forwarded.getOrElse {
-              range(super.derivedSelect(tp, preLo).lowerBound, super.derivedSelect(tp, preHi).upperBound)
+              range(super.derivedSelect(tp, preLo) /*.lowerBound*/, super.derivedSelect(tp, preHi) /*.upperBound*/ )
             }
           case _ =>
             super.derivedSelect(tp, pre) match {
-              case tp: WildcardTypeBounds => range(tp.bounds.low, tp.bounds.high)
-              case tp                     => tp
+              //case tp: WildcardTypeBounds => range(tp.bounds.low, tp.bounds.high)
```
bishabosha, line 301: is this now illegal or temporary removal?

tasty-query/shared/src/main/scala/tastyquery/TypeMaps.scala
```diff
@@ -301,12 +294,12 @@ private[tastyquery] object TypeMaps {
               case Some(sym: ClassTypeParamSymbol) => expandParam(sym, preHi)
               case _                               => tryWiden(tp, preHi)
             forwarded.getOrElse {
-              range(super.derivedSelect(tp, preLo).lowerBound, super.derivedSelect(tp, preHi).upperBound)
+              range(super.derivedSelect(tp, preLo) /*.lowerBound*/, super.derivedSelect(tp, preHi) /*.upperBound*/ )
```
bishabosha, line 297: these are no longer needed?

tasty-query/shared/src/main/scala/tastyquery/Types.scala
```diff
@@ -38,50 +38,71 @@ import tastyquery.Variances.*
...
-  *        +- RecType              recursive type that introduces a recursive `this` binding
-  *        +- RecThis              recursive reference to the `this` of a `RecType`
-  *        +- SuperType            `super[mix]` (typically used as `prefix` of `NamedType`s)
+  *  +- TypeOrMethodic               a type or a methodic type (e.g., the declared type of a term symbol)
+  *      |
+  *      +- MethodicType             the type of a method with at least one parameter list
+  *      |   +- MethodType           `(termParams): resultType`
+  *      |   +- PolyType             `[TypeParams]: resultType`
+  *      |
+  *      +  Type
```
bishabosha, line 54:
```suggestion
  *      +- Type
```

bishabosha:
rebase needed :)

tasty-query/shared/src/main/scala/tastyquery/Definitions.scala
sjrd: It's not supposed to. Unfortunately, the two call sites of this method translate from `AppliedType`s of `Seq`s or `Array`s, so statically we don't know that it's not a wildcard. Since the rest of the system would not break with a wildcard here, I consider it safer to allow it.

tasty-query/shared/src/main/scala/tastyquery/TypeMaps.scala
sjrd: This is a pretty gray area. I still don't understand what's happening with those `range`s. While the refactoring suggests that a wildcard cannot appear here, I'd rather keep a trace that we *used to* deal with wildcards here, in case it comes to bite back in the future.

That said, I turned that into an explicit textual comment to make it clearer.

tasty-query/shared/src/main/scala/tastyquery/TypeMaps.scala
sjrd: Same here.

bishabosha (approved):

PR #305, opened 2023-06-09
Implement the relationship between TupleN classes and generic tuples.

sjrd:


tasty-query/shared/src/main/scala/tastyquery/Definitions.scala
```diff
@@ -70,6 +70,19 @@ final class Definitions private[tastyquery] (ctx: Context, rootPackage: PackageS
 
   val ThrowableType: TypeRef = TypeRef(javaLangPackage.packageRef, typeName("Throwable"))
 
+  val TupleType: TypeRef = TypeRef(scalaPackage.packageRef, typeName("Tuple"))
+
+  val EmptyTupleType: TypeRef =
```
bishabosha, line 75: shouldn't this be the `TermRef` to the module?

tasty-query/shared/src/main/scala/tastyquery/Subtyping.scala
```diff
@@ -294,8 +294,8 @@ private[tastyquery] object Subtyping:
         else
           tycon2.optSymbol match
             case Some(cls2: ClassSymbol) =>
-              // TODO Handle generic tuple <: TupleN
-              level3WithBaseType(tp1, tp2, cls2)
+              (defn.isTupleNClass(cls2) && isSubtype(tp1, defn.GenericTupleTypeOf(tp2.args)))
```
bishabosha, line 297: potential optimisation: check tp1 is `*:` first before creating the generic tuple type?

sjrd: That's not trivial at all, because we know nothing about `tp1` at this point. It could be a complicated type alias that eventually becomes a `*:`.

sjrd:
Rebased and updated.

bishabosha:
MiMA check doesn't like the change to the `EmptyTuple` in names

bishabosha (approved):

PR #306, opened 2023-06-14
Fixes for applying tasty-mima across Scala 2 and Scala 3

sjrd:
@nicolasstucki Do you want to review this one? There's not everything yet, notably the `InternalError`s, but it's already something.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
@@ -519,7 +519,16 @@ private[pickles] class PickleReader {
         else if (args.nonEmpty) tycon.safeAppliedTo(EtaExpandIfHK(sym.typeParams, args.map(translateTempPoly)))
         else if (sym.typeParams.nonEmpty) tycon.EtaExpand(sym.typeParams)
         else tycon*/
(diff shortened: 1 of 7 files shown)


tasty-query/shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
```diff
@@ -692,6 +736,8 @@ private[pickles] class PickleReader {
 
 private[reader] object PickleReader {
 
+  private val MangledDefaultGetterNameRegex = "^(.+)\\$default\\$([0-9]+)$".r
```
nicolasstucki, line 739: Does this work with default constructor arguments?  Maybe add a test for those.

nicolasstucki, line 739: Does this work with default constructor arguments?  Maybe add a test for those.

nicolasstucki (commented):
Otherwise LGTM

tasty-query/shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
sjrd: It works. I added a test :)

PR #314, opened 2023-06-23
Include the isStable flag in `TermMemberResult`.

sjrd:
So that selecting from a `val` term refinement indeed gives a stable `TermRef`.

This also brings term member resolution closer to the spec.

### DIFF
--- a/tasty-query/shared/src/test/scala/tastyquery/SubtypingSuite.scala
+++ b/tasty-query/shared/src/test/scala/tastyquery/SubtypingSuite.scala
@@ -716,8 +716,8 @@ class SubtypingSuite extends UnrestrictedUnpicklingSuite:
       val polyHigh = TypeLambda(List(typeName("X")), List(defn.NothingAnyBounds), high)
       TypeRefinement(SimplePathsClass.appliedRef, typeName(name), RealTypeBounds(polyLow, polyHigh))
 
-    def refineTerm(name: String, tpe: TypeOrMethodic): Type =
-      TermRefinement(SimplePathsClass.appliedRef, isStable = false, termName(name), tpe)
+    def refineTerm(name: String, tpe: TypeOrMethodic, isStable: Boolean = false): Type =
+      TermRefinement(SimplePathsClass.appliedRef, isStable, termName(name), tpe)
 
     // type refinement - exists in class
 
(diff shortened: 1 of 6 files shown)


tasty-query/shared/src/test/scala/tastyquery/SubtypingSuite.scala
```diff
@@ -794,6 +794,35 @@ class SubtypingSuite extends UnrestrictedUnpicklingSuite:
...
+    assertEquiv(
+      refineTerm("abstractTerm", defn.StringType, isStable = true),
+      refineTerm("abstractTerm", defn.StringType, isStable = true)
+    )
+      .withRef[SimplePaths { val abstractTerm: String }, SimplePaths { val abstractTerm: String }]
+    assertStrictSubtype(refineTerm("abstractTerm", defn.StringType, isStable = true), SimplePathsClass.appliedRef)
+      .withRef[SimplePaths { val abstractTerm: String }, SimplePaths]
+
+    // no withRef on the following because dotc does not take stability into account, apparently
+    assertStrictSubtype(
```
bishabosha, line 806: if there is a difference to dotc isn't that incorrect?

bishabosha (approved):

tasty-query/shared/src/test/scala/tastyquery/SubtypingSuite.scala
sjrd: I mean, yes, dotc is incorrect. :laughing: It's unsound. Scala 2 is right:
```scala
$ cs launch scala:2.13.11
Welcome to Scala 2.13.11 (OpenJDK 64-Bit Server VM, Java 1.8.0_372).
Type in expressions for evaluation. Or try :help.

scala> def foo(x: Any { def foo: Int }): Any { val foo: Int } = x
                                                                ^
       error: type mismatch;
        found   : Any{def foo: Int}
        required: Any{val foo: Int}

scala> 
$ cs launch scala:3.3.0
Welcome to Scala 3.3.0 (1.8.0_372, Java OpenJDK 64-Bit Server VM).
Type in expressions for evaluation. Or try :help.
                                                                                                    
scala> def foo(x: Any { def foo: Int }): Any { val foo: Int } = x
def foo(x: Any{def foo: Int}): Any{val foo: Int}
                                                                                                    
scala> 
```

sjrd: I reported https://github.com/lampepfl/dotty/issues/18047

PR #318, opened 2023-07-04
Remove TermRef.symbol and use optSymbol instead.

sjrd:
In fact, it is possible to select a term that does not have a symbol, as a type. It is not allowed as the type of an `Ident` or `Select`, however.


sjrd:
Hum, I have to double-check whether this is actually needed and when. I thought they could participate in paths, but it does not seem to be the case.

sjrd:
Really closing this now. The doubts I still had about needing this were addressed in #362.

PR #325, opened 2023-07-13
Harmonize isSubX methods to start X with an uppercase.

sjrd:
Concretely, rename isSubclass, isSubtype and isSubprefix to isSubClass, isSubType and isSubPrefix. It makes it more consistent with isSubTypeOrMethodic and isSameType.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Types.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Types.scala
@@ -355,8 +355,8 @@ object Types {
       case self: Type =>
         that match
           case that: Type =>
-            if self.isSubtype(that) then self
-            else if that.isSubtype(self) then that
+            if self.isSubType(that) then self
+            else if that.isSubType(self) then that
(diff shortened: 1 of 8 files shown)


tasty-query/shared/src/main/scala/tastyquery/Types.scala
```diff
@@ -472,8 +472,12 @@ object Types {
   sealed abstract class Type extends TypeOrMethodic with NonEmptyPrefix with TypeOrWildcard:
     private[tastyquery] type ThisTypeMappableType = Type
 
+    @deprecated("use isSubType instead", since = "0.9.9")
```
bishabosha, line 475:
```suggestion
    @deprecated("use isSubType instead", since = "0.9.0")
```

sjrd: Oops. Fixed.

bishabosha (approved):

PR #326, opened 2023-07-13
Remove useless contexts

### LINKED ISSUE #273: Identify/document which APIs are safe to use without a full classpath
Have a list, or clearly document in the docs, of what methods from the API one can safely use without providing the full classpath. These would be mostly "syntactic" APIs, obviously.

Coming from the discussion at the Scala Tooling Summit last week.

sjrd:
As a by-product, this PR closes #273. APIs that are safe to use without a full classpath do not take a `using Context` as parameter.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
@@ -1619,7 +1619,7 @@ object Symbols {
     private val myDeclarations = mutable.HashMap[Name, Symbol]()
 
     // Cache fields
-    val packageRef: PackageRef = PackageRef(this: @unchecked)
+    val packageRef: PackageRef = new PackageRef(this)
     private var myAllPackageObjectDecls: List[ClassSymbol] | Null = null
 
     this.withFlags(EmptyFlagSet, None)
--- a/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
@@ -103,7 +103,7 @@ object Symbols {
       myTree = Some(t)
       this
 
-    final def tree(using Context): Option[DefiningTreeType] =
+    final def tree: Option[DefiningTreeType] =
       myTree
 
     private[tastyquery] final def withFlags(flags: FlagSet, privateWithin: Option[DeclaringSymbol]): this.type =
@@ -401,7 +401,7 @@ object Symbols {
       myDeclaredType = tpe
       this
 
-    def declaredType(using Context): TypeOrMethodic =
+    def declaredType: TypeOrMethodic =
       val local = myDeclaredType
       if local != null then local
       else throw new IllegalStateException(s"$this was not assigned a declared type")
@@ -496,7 +496,7 @@ object Symbols {
     private def isConstructor: Boolean =
       owner.isClass && isMethod && name == nme.Constructor
 
-    private[tastyquery] final def needsSignature(using Context): Boolean =
+    private[tastyquery] final def needsSignature: Boolean =
       declaredType.isInstanceOf[MethodicType]
 
     final def signature(using Context): Signature =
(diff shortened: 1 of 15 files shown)


tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -684,7 +688,11 @@ object Symbols {
     def variance(using Context): Variance =
       declaredVariance
 
-    final def typeRef(using Context): TypeRef = TypeRef(ThisType(owner.typeRef), this)
+    @deprecated("use localRef instead", since = "0.9.0")
+    final def typeRef: TypeRef = localRef
```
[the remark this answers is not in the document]
sjrd: We're in a backward-incompatible release anyway, so it doesn't matter. If we took that argument, we would need a deprecated overload of every method for which we removed `(using Context)`.

tasty-query/shared/src/main/scala/tastyquery/reader/tasties/TreeUnpickler.scala
```diff
@@ -1064,6 +1065,24 @@ private[tasties] class TreeUnpickler private (
...
+    *
+    * In a full, correct classpath, `readPackageRef()` will always return a
+    * `PackageRef`. However, in an incomplete or incorrect classpath, this
+    * method may return a `TermRef` if the target package does not exist.
+    *
+    * An alternative would be to create missing packages on the fly, but that
+    * would not be consistent with `Trees.Select.tpe` and
+    * `Trees.TermRefTypeTree.toType`.
+    */
+  private def readPackageRef(): PackageRef | TermRef =
```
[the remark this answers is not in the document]
bishabosha: or create the package here?

tasty-query/shared/src/main/scala/tastyquery/Trees.scala
```diff
@@ -102,11 +102,34 @@ object Trees {
...
 
     def withSpan(span: Span): TermTree
 
-    /** Compute the actual term of this tree. */
+    /** Compute the prefix represented by this tree.
+      *
+      * The default implementation throws an `InvalidProgramStructureException`
+      * to indicate that this term tree is not a valid type prefix.
+      */
+    protected def computeAsTypePrefix: NonEmptyPrefix =
```
bishabosha, line 115: why this and not create a new PrefixTermTree?

sjrd: I considered that but it doesn't work. A `Select` has *sometimes* a valid `toTypePrefix`, sometimes not, depending on whether its `qualifier` has a valid `toTypePrefix` or not. There are similar problems with other kinds of trees. We would have needed a parallel hierarchy of these things, which wouldn't have played well with the unpickler anyway. So all in all, this is the only viable design.

tasty-query/shared/src/main/scala/tastyquery/reader/tasties/TreeUnpickler.scala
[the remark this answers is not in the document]
sjrd: If we create package symbols on-demand, even if marked as synthetic, we start having big problems if a synthetic package symbol clashes with an actual, non-package term symbol. Creating all package symbols first, then none later, is the only way to prevent non-determinism in that space, AFAICT. That means not having symbols for should-be-`PackageRef`s that don't actually exist.



> this is arguably pretty bad because syntactically its a package ref, and we wanted to make syntax non-context dependent. what about making PackageRef have an optional symbol at construction? (so still accessible without context)

Sort of, but we have precedent as far as package refs are concerned. When unpickling Scala 2 pickles, for example, the corresponding encoding already behaves in what this PR brings to .tasty files. Indeed, in Scala 2, there is only "external symbol ref" with a path that starts from root. We eagerly follow packages that exist to create `PackageRef`s, and once we don't find a package anymore, we proceed with `TermRef`s. This is not really different.

In practice, error messages will be decent, as we will anyway end up with `MemberNotFoundException`s when we try to access the dangling `TermRef`s'symbols.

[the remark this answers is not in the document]
bishabosha: ok, that is reasonable, (user should supply the correct classpath anyway)

tasty-query/shared/src/main/scala/tastyquery/Trees.scala
bishabosha: when would a qualifier not be a valid prefix?

bishabosha: oh yeah, true

bishabosha: so how does someone use this API, call it and hope there's no Apply in the prefix?

bishabosha: Ok it seems to only be used to calculate the prefix of a singleton type tree, which should be properly constructed - int hat case would you consider making it `private[tastyquery]`?

sjrd: OK, I made it `private[Trees]`. And since it is then guaranteed to be called at most once per `TermTree`, I removed the cache field.

bishabosha (approved):

PR #327, opened 2023-07-17
Cleanups

sjrd:


tasty-query/shared/src/main/scala/tastyquery/Types.scala
```diff
@@ -484,20 +495,27 @@ object Types {
...
+        tpe.underlyingOrMethodic match
+          case underlying: Type => underlying.classSymbol
+          case _: MethodicType  => None
       case tpe: TypeProxy =>
         tpe.superType.classSymbol
       case tpe: TypeLambda =>
         // apparently we need this :(
         tpe.resultType.classSymbol
-      case _ =>
+      case _: NothingType | _: AnyKindType | _: OrType | _: AndType =>
```
bishabosha, line 514: I would like to see that `(Array[String] with Array[String]) with Array[String]` in a scala 2 source still erases correctly

bishabosha: it looks like these parents will be flattened though so it should be fine

bishabosha (approved):

tasty-query/shared/src/main/scala/tastyquery/Types.scala
sjrd: We don't compile any test source using Scala 2, so it's a bit difficult to this from source. I added a direct unit test of erasure instead.

PR #330, opened 2023-07-19
Make Scala 2 getters appear as fields, to be consistent with Scala 3.

sjrd:
And hide the actual fields, whose name ends in ' '.


adpi2 (approved):
Thanks

PR #340, opened 2023-08-10
Replace Spans by SourcePositions.

sjrd:
Unlike spans, SourcePositions keep a reference to their SourceFile. The latter stores the sizes of its lines, which allows it to reconstruct line and column numbers from offsets.


tasty-query/shared/src/main/scala/tastyquery/SourceFile.scala
```diff
@@ -0,0 +1,55 @@
...
+    else throw UnsupportedOperationException(s"Source file $path does not have line/column information")
+
+  private def lineToOffset(line: Int): Int =
+    ensureLineStartOffsets()(line)
+
+  private[tastyquery] def offsetToLine(offset: Int): Int =
+    val startOffsets = ensureLineStartOffsets()
+    val p = Arrays.binarySearch(startOffsets, offset)
+    if p >= 0 then p
+    else (-p - 1) - 1 // (-p - 1) is the insertion point, from which we still need to subtract 1 to get the line
```
adpi2, line 42: That's funky!

tasty-query/shared/src/main/scala/tastyquery/SourcePosition.scala
```diff
@@ -0,0 +1,74 @@
...
+  def endOffset: Int = span.end
+
+  /** The line number of the end of this position. */
+  def endLine: Int = sourceFile.offsetToLine(endOffset)
+
+  /** The column number of the end of this position. */
+  def endColumn: Int = sourceFile.offsetToColumn(endOffset)
+
+  /** Is this position synthetic? */
+  def isSynthetic: Boolean = span.isSynthetic
```
adpi2, line 56: What's a synthetic span?

adpi2 (approved):
Formidable!

tasty-query/shared/src/main/scala/tastyquery/SourcePosition.scala
sjrd: I added a comment. Is that clearer?

PR #347, opened 2023-09-08
Introduce a more restrictive `ReaderContext` inside `reader.*`.

sjrd:
It statically makes sure that we never access anything that requires reading another file.

This change surfaced a couple places where there could potentially be external reading, and which are now a bit more complicated to fit the static restriction. The most prominent example is `TreeUnpickler.readWithin`, which also suggested that we decouple reading the `privateWithin` from the flags (to only do it after `createSymbols()` is over).

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/reader/classfiles/ClassfileParser.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/reader/classfiles/ClassfileParser.scala
@@ -13,6 +13,8 @@ import tastyquery.SourceLanguage
 import tastyquery.Symbols.*
 import tastyquery.Types.*
 
+import tastyquery.reader.ReaderContext
+import tastyquery.reader.ReaderContext.rctx
 import tastyquery.reader.pickles.{Unpickler, PickleReader}
 
 import ClassfileReader.*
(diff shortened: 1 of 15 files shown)


tasty-query/shared/src/main/scala/tastyquery/reader/classfiles/ClassfileParser.scala
```diff
@@ -38,29 +39,32 @@ private[reader] object ClassfileParser {
     private val refs = mutable.HashMap.empty[SimpleName, TypeRef]
     private val staticrefs = mutable.HashMap.empty[SimpleName, TermRef]
 
-    private def computeRef(binaryName: SimpleName, isStatic: Boolean)(using Context, InnerClasses): NamedType =
+    private def computeRef(binaryName: SimpleName, isStatic: Boolean)(
+      using ctx: ReaderContext,
```
bishabosha, line 43: this could be cleaned up by creating a summoner method for `ReaderContext`

tasty-query/shared/src/main/scala/tastyquery/reader/classfiles/JavaSignatures.scala
```diff
@@ -199,10 +201,10 @@ private[classfiles] object JavaSignatures:
         expect(':')
         referenceTypeSignature(env) match
           case Some(tpe) => tpe
-          case _         => defn.FromJavaObjectType
+          case _         => ctx.FromJavaObjectType
       val interfaceBounds = readWhile(':', referenceType(env))
       if env.withAddedParam(tname) then emptyTypeBounds // shortcut as we will throw away the bounds
-      else RealTypeBounds(defn.NothingType, interfaceBounds.foldLeft(classBound)(AndType(_, _)))
+      else RealTypeBounds(ctx.NothingType, interfaceBounds.foldLeft(classBound)(AndType(_, _)))
```
bishabosha, line 207: I wonder about readability, what do you think about using `rctx`?

bishabosha (approved):
other than stylistic comments

sjrd:
Updated with an `rctx` summoner.

bishabosha (approved):

PR #350, opened 2023-09-25
Fix reading of source file information.

sjrd:
I had completely misunderstood the format of source file information. Previously, I thought they applied to a range of position information in the positions section of the TASTy file. In fact, they apply to sub-*trees* rooted in any given tree.

I locally verified that it fixes the following test when adding the compiler to the test classpath:
```scala
  testWithContext("enum position") {
    val BindingPrecClass = ctx.findStaticClass("dotty.tools.dotc.typer.Typer.BindingPrec")

    def desc(pos: SourcePosition): String =
      "${pos.sourceFile.name}:${pos.startLine}:${pos.startColumn}-${pos.pointLine}:${pos.pointColumn}-${pos.endLine}:${pos.ndColumn}"

    val cd = BindingPrecClass.tree.get
    println(cd.pos)
    println(desc(cd.pos))
    println(cd.rhs.pos)
    println(desc(cd.rhs.pos))
  }
```
However, it seems difficult to reproduce without complicate source files depending on other complicated source files through inlining, so I could not reproduce it as a unit test.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -117,6 +117,7 @@ lazy val tastyQuery =
 
           // private[tastyquery], not an issue
           ProblemFilters.exclude[IncompatibleMethTypeProblem]("tastyquery.Symbols#ClassSymbol.createRefinedClassSymbol"),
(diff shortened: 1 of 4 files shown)


adpi2 (approved):
Really nice! This fixed a lot of errors in the better stack-trace tests.

PR #351, opened 2023-09-27
Automatically fill in positions of trees based on their children.

sjrd:
When pickling, dotc does not store the span of trees for which it is the same as the union of the spans of its children.

In order to recover those, we now have an internal "auto" mode of `SourcePosition`, which stores the `SourceFile` but with `NoSpan`. When building a `Tree`, if it is given an "auto" position, we compute the span based on its children. We mimick the formula used by dotc to do so, but only upward (dotc also does some weird mutating downward propagation, but AFAICT it will always store the spans of leaf trees, so we don't need that).


adpi2:
Thanks I'll check that in the afternoon!

adpi2 (approved):
It fixed many of the not-found symbols, including the last 2 classes!

PR #366, opened 2023-10-20
Redesign Names.

sjrd:


tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -1689,6 +1696,12 @@ object Symbols {
     this.withFlags(EmptyFlagSet, None)
     this.setAnnotations(Nil)
 
+    private lazy val _fullName: FullyQualifiedName =
+      if owner == null || name == nme.EmptyPackageName then FullyQualifiedName.rootPackageName
```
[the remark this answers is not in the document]
bishabosha: I can see its just a refactoring, so not really important

tasty-query/shared/src/main/scala/tastyquery/reader/Loaders.scala
```diff
@@ -18,13 +18,12 @@ import tastyquery.reader.tasties.TastyUnpickler
...
-  class MissingTopLevelTasty(root: Loader.Root) extends Exception(s"Missing TASTy for ${root.fullName}")
-  class UnexpectedTasty(root: Loader.Root) extends Exception(s"Unexpected TASTy for ${root.fullName}")
+  class MissingTopLevelTasty(root: Loader.Root) extends Exception(s"Missing TASTy for $root")
+  class UnexpectedTasty(root: Loader.Root) extends Exception(s"Unexpected TASTy for $root")
 
   object Loader:
     private[Loaders] case class Root private[Loaders] (pkg: PackageSymbol, rootName: SimpleName):
-      def fullName: FullyQualifiedName =
-        pkg.fullName.select(rootName)
+      override def toString(): String = pkg.displayFullName + "." + rootName.toString
```
bishabosha, line 26: this will now print `<empty>.Foo` rather than `Foo`

sjrd: Fixed.

bishabosha (approved):

PR #378, opened 2023-11-02
Fix reading VALsym's without owner in the Scala 2 unpickler.

sjrd:
This was discovered by checking sbt-tasty-mima itself with TASTy-MiMa.


bishabosha (approved):
would be nice for someone to add pretty printing of scala 2 pickle tags

PR #381, opened 2023-11-08
Fix #380: Give the correct type to the synthetic ctor of j.l.Enum.

### LINKED ISSUE #380: Computing type of `new java.lang.Enum[Foo]()` throws `NonMethodReferenceException`
```scala
enum Foo extends java.lang.Enum[Foo]:
  case Bar
```

Somewhere inside the tree of `Foo` there is an `Apply` tree: `new java.lang.Enum[Foo]()`. If I compute the type of this tree it throws `NonMethodReferenceException` because the type of the `fun` is not a method type.

sjrd:


adpi2 (approved):
Thanks!

PR #386, opened 2023-11-13
Remove `myParentsInit`/`withParentsDelayed`; use the `tree` instead.

sjrd:
The information stored in `myParentsInit` can be found in the `tree` of the symbol, so we directly get it from there. This avoids storing a lambda with delayed evaluation in the symbols, which is more in line with our model of only storing *data*, not *computations*.

Note that we cannot do this for the other information we store in symbols, because they can come from Scala 2 pickles and Java class files, which have no trees. `myParentsInit` is the odd one in that it is already a TASTy-specific alternative to `myParents`. `myParents` is directly filled in for Scala 2 and Java.


bishabosha (approved):

bishabosha:
nice simplification

PR #389, opened 2023-11-16
Read Java annotations.

sjrd:
Bonus: read Java parameter names.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/reader/classfiles/ClassfileParser.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/reader/classfiles/ClassfileParser.scala
@@ -1,23 +1,28 @@
 package tastyquery.reader.classfiles
 
+import scala.annotation.switch
+
 import scala.collection.mutable
 
-import tastyquery.Annotations.Annotation as TQAnnotation
+import tastyquery.Annotations.*
 import tastyquery.Classpaths.*
 import tastyquery.Contexts.*
+import tastyquery.Constants.*
 import tastyquery.Exceptions.*
 import tastyquery.Flags
 import tastyquery.Flags.*
 import tastyquery.Names.*
 import tastyquery.SourceLanguage
+import tastyquery.SourcePosition
 import tastyquery.Symbols.*
+import tastyquery.Trees.*
 import tastyquery.Types.*
 
 import tastyquery.reader.ReaderContext
 import tastyquery.reader.ReaderContext.rctx
 import tastyquery.reader.pickles.{Unpickler, PickleReader}
 
-import ClassfileReader.*
+import ClassfileReader.{Annotation as CFAnnotation, *}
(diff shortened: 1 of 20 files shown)


tasty-query/shared/src/main/scala/tastyquery/reader/classfiles/ClassfileParser.scala
```diff
@@ -324,6 +333,88 @@ private[reader] object ClassfileParser {
...
+        val binaryName = descriptor.name match
+          case s"L$binaryName;" => binaryName
+          case other            => throw ClassfileFormatException(s"unexpected non-class field descriptor: $other")
+        val enumClassStaticRef = resolver.resolveStatic(termName(binaryName))
+        val constRef = TermRef(enumClassStaticRef, constName)
+        Ident(constName)(constRef)(pos)
+
+      case AnnotationValue.ClassConst(descriptor) =>
+        val classType = JavaSignatures.parseReturnDescriptor(descriptor.name)
+        Literal(Constant(classType))(pos)
```
bishabosha, line 474: documentation of Constant should mention that this can be a raw class type

bishabosha: maybe add a test for a generic external class

tasty-query/shared/src/main/scala/tastyquery/reader/classfiles/JavaSignatures.scala
```diff
@@ -38,10 +38,51 @@ private[classfiles] object JavaSignatures:
       superRef :: interfaces.map(classRef).toList
   end parseSupers
 
+  def parseReturnDescriptor(descriptor: String)(using ReaderContext, InnerClasses, Resolver): Type =
```
bishabosha, line 41: this is also implemented by the local method `result` inside of `parseSignature`, so I wonder if it can be reused with refactoring?

bishabosha (approved):

tasty-query/shared/src/main/scala/tastyquery/reader/classfiles/ClassfileParser.scala
sjrd: Added documentation + test.

tasty-query/shared/src/main/scala/tastyquery/reader/classfiles/JavaSignatures.scala
sjrd: Yes, I know. It's not that easy, however, because the more capable `parseSignature` expects a `member: TermOrTypeSymbol` that the signature attaches to, which we have no business having here. Factoring this out would require significant refactoring of the whole parsing, so I would rather not do that here (at least not at this point).

bishabosha (approved):

PR #390, opened 2023-11-16
Add methods to detect parameters with default values.

sjrd:
`TermSymbol.isParamWithDefault` returns true on term parameter symbols that have a default value.

`TermSymbol.hasParamsWithDefault` returns true on method symbols that have at least one parameter for which `isParamWithDefault` is true.


tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -583,6 +583,17 @@ object Symbols {
     /** Is this symbol an exporter generated by an `export` statement? */
     final def isExport: Boolean = flags.is(Exported)
 
+    /** Is this symbol a method with at least one parameter with a default value? */
+    final def hasParamsWithDefault: Boolean =
```
adpi2, line 587: Should we use a singular here, as in "has at least one param with default" ?

```suggestion
    final def hasParamWithDefault: Boolean =
```

adpi2 (approved):
Apart from the small suggestion, it looks good to me.

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
sjrd: Indeed, that's probably better. Changed.

PR #391, opened 2023-11-17
Make Constants.Constant final.

sjrd:
**Potentially controversial.**

This is technically a compatibility breach, both at the binary and TASTy level. However, I don't see how anyone could have meaningfully extended `Constant` (also we have 2 users so far).


bishabosha (approved):

bishabosha:
Hypothetically if there are custom trees allowed maybe there could be custom constants

adpi2 (approved):

PR #392, opened 2023-11-17
Remove useless flags.

sjrd:
`VarianceFlags` was dead code.

`TypeParameter` and `ClassTypeParam` are redundant with the actual classes of type symbols.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Flags.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Flags.scala
@@ -78,16 +78,10 @@ private[tastyquery] object Flags:
   val SignaturePolymorphic: Flag = newFlag("SignaturePolymorphic")
   val SuperParamAlias: Flag = newFlag("SuperParamAlias")
   val Static: Flag = newFlag("Static")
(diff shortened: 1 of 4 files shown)


tasty-query/shared/src/main/scala/tastyquery/Flags.scala
```diff
@@ -78,16 +78,10 @@ private[tastyquery] object Flags:
   val SignaturePolymorphic: Flag = newFlag("SignaturePolymorphic")
   val SuperParamAlias: Flag = newFlag("SuperParamAlias")
   val Static: Flag = newFlag("Static")
-  private[tastyquery] val StableRealizable: Flag = newFlag("StableRealizable")
+  val StableRealizable: Flag = newFlag("StableRealizable")
```
bishabosha, line 81: im sure this was private because it was flaky if it actually appears in tasty

sjrd: The whole `Flags` is private now. 😉

bishabosha (approved):

PR #393, opened 2023-11-19
Add a basic printer for trees, with a method Tree.showBasic.

sjrd:
I only wrote tests for the weirdest forms of trees.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Printers.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Printers.scala
@@ -75,10 +75,7 @@ private[tastyquery] object Printers:
       case tpe: AppliedType =>
         print(tpe.tycon)
         print("[")
-        print(tpe.args.head)
-        for arg <- tpe.args.tail do
-          print(", ")
-          print(arg)
+        printCommaSeparatedList(tpe.args)(print(_))
         print("]")
       case tpe: ByNameType =>
         print("=> ")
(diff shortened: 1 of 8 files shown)


tasty-query/shared/src/main/scala/tastyquery/Printers.scala
```diff
@@ -204,6 +192,545 @@ private[tastyquery] object Printers:
...
+          printElem(elem)
+    end printCommaSeparatedList
+
+    def printBlock[A](elems: List[A])(printElem: A => Unit): Unit =
+      if elems.isEmpty then print(" {}")
+      else
+        print(" { ")
+        printElem(elems.head)
+        for elem <- elems.tail do
+          print("; ")
```
adpi2, line 729: Would it be more convenient to have a line break here?

```suggestion
          print("\n")
```

bishabosha: I guess then there should be a stack of indentation levels to respect

sjrd: Yes, if we put newlines we must deal with indentation. My thought was that it would be left for a subclass that can surgically override `printBlock`.

adpi2: I don't think we need any indentation since we have "{ }" around all blocks.

sjrd:
I added a commit with `showMultiline` and multiline printers. Those take care of indentation.

adpi2 (approved):

PR #395, opened 2023-11-22
Work around lampepfl/dotty#19019: patch the SelfDef of inner module classes.

sjrd:
Re-create the correct self type which is a term reference to the accompanying lazy val.

### DIFF
--- a/test-sources/src/main/scala/simple_trees/ObjectWithSelf.scala
+++ b/test-sources/src/main/scala/simple_trees/ObjectWithSelf.scala
@@ -1,0 +1,43 @@
+package simple_trees
+
+object ObjectWithSelf:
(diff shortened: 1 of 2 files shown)


test-sources/src/main/scala/simple_trees/ObjectWithSelf.scala
adpi2, line 1: Do we check this tree in some test?

sjrd: Not the tree itself, but we do check, through `WholeClasspathSuite`, that every `.symbol` and `.tpe` in the entire classpath (including test-sources) can be successfully resolved. So that tests the original issue.

adpi2 (approved):

PR #408, opened 2023-11-30
Fix #405: Completely overhaul erasure of value classes.

### LINKED ISSUE #405: Wrong erasure of parametric value classes
```scala
package example

class Value[T](v: T) extends AnyVal

class C[T <: Int]:
  def m(x: Value[T]): Unit = ???
```

`ErasedTypeRef.erase` on the type of `x` returns `Object` instead of `Int`

_Originally posted by @adpi2 in https://github.com/scalacenter/tasty-query/issues/403#issuecomment-1831874296_

sjrd:
Including when they contain arrays or are contained in arrays.


adpi2 (approved):
Tested on the binary decoder. It works well.

PR #411, opened 2023-12-04
Cleanup erasure: do not delay erasing special classes.

sjrd:
Now that generic arrays are dealt with ahead of time, using `isGenericArrayElement`, the delay is not necessary anymore.

This is a bit cleaner, and also allows us to directly use `cls.erasure` everywhere, which will avoid allocating temporary `ClassRef`s in most situations.


adpi2 (approved):
It does not break any of my tests so I guess it's good

PR #416, opened 2023-12-04
Fix #412: Protect the state of packages and loaders from exceptions.

### LINKED ISSUE #412: Exception in `TreeUnpickler` corrupts entries in `PackageSymbols`
Reproduction steps:

1. In `"dev.zio" %% "zio" % "2.0.16"`
2. Call `declarations` on `PackageSymbol` of `zio`. It throws:
```
tastyquery.Exceptions$TastyFormatException: Unexpected type tag APPLY at address Addr(324) in file ZLayerCompanionVersionSpecific.class
```
4.Try resolve member `package` on `PackageSymbol` of `zio`. It throws a 
```
tastyquery.Exceptions$MemberNotFoundException: Member package not found in PackageRef(zio)
```

sjrd:
At the loader level:
Only remove package data entries *after* successful loading.

At the package symbol level:
While loading new symbols from the loader, all new declarations are put in a `pendingDeclarations` safe zone. They are committed to `myDeclarations` only after the whole operation successfully complets.

This way, if a failure happens while reading one root, it does not corrupt the other roots in the same package.

---

Locally tested with
```scala
    libraryDependencies += "dev.zio" %% "zio-test" % "2.0.16",
```
and the following `TypeSuite` test:
```scala
  testWithContext("issue-412") {
    val zioPackage = ctx.findPackage("zio")

    // Cause the initial failure
    intercept[TastyFormatException](zioPackage.declarations)

    val decl1 = zioPackage.findDecl(typeName("ZInputStream")) // before problematic file
    assert(decl1.isClass)

    val decl2 = zioPackage.findDecl(typeName("ZOutputStream")) // after problematic file
    assert(decl2.isClass)

    val decl3 = zioPackage.findDecl(termName("package")).asTerm // original issue
    assert(decl3.isModuleVal)

    // Trying to read the problematic file still reports the same exception
    intercept[TastyFormatException](zioPackage.findDecl(termName("ZLayerCompanionVersionSpecific")))

    // Trying to load everything again still reports the same exception
    intercept[TastyFormatException](zioPackage.declarations)
  }
```

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
@@ -1801,6 +1801,8 @@ object Symbols {
     // DeclaringSymbol-related fields
     private var rootsInitialized: Boolean = false
     private val myDeclarations = mutable.HashMap[UnsignedName, Symbol]()
+    private val pendingDeclarations = mutable.HashMap[UnsignedName, Symbol]()
+    private var isLoadingNewRoots: Boolean = false
 
     // Cache fields
     val packageRef: PackageRef = new PackageRef(this)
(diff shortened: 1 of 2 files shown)


sjrd:
Hum this is still broken. I'll fix it tomorrow.

sjrd:
OK I believe this really good, now. I enhanced the test in the PR description.

adpi2 (approved):

adpi2:
Actually there is a little regression in this. Before it would add all the symbols in the file up to the exception, but now it does not add any. Is this the desired behavior?

```
object Foo

class Foo
```

I cannot find the `object Foo` anymore even if the exception is in the TASTy part of `class Foo`.

sjrd:
Yes, that's expected. A companion and its class may have symbolic links between each other. If we throw away one but keep the other, it may contain and expose references to `Symbol`s that were never properly completed.

PR #420, opened 2023-12-12
Do not crash when reading a Scala 2 `macro` definition.

sjrd:
It contains an annotation with a subtree that is an `IDENTtree` without prefix nor symbol, with the name `macro`. That is then `Apply`ed and `TypeApply`ed in ways that make no semantic sense.

We define a special symbol in `Definitions`, which we attach to such fake identifiers.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,6 +5,7 @@ import org.scalajs.jsenv.nodejs.NodeJSEnv
 
 val usedScalaCompiler = "3.3.1"
 val usedTastyRelease = usedScalaCompiler
+val scala2Version = "2.13.12"
 
 val SourceDeps = config("sourcedeps").hide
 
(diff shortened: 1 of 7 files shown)


adpi2 (commented):
Don't you have some tests for Scala 2?

sjrd:
Only for things that we can find in the standard library. We don't have a setup for other weird stuff that we only find elsewhere, unfortunately.

adpi2:
What about adding a Scala 2 test project like this one?
https://github.com/scalacenter/tasty-query/blob/917d4d6111d49820b070eb784ba2eb601ef6937d/build.sbt#L64C1-L72C4

sjrd:
All right, I added a `scala2TestSources` project to be able to test targeted Scala 2 stuff, and I added a specific test for the fix in this PR.

adpi2 (approved):

PR #425, opened 2023-12-18
Work around https://github.com/lampepfl/dotty/issues/19237

sjrd:
It does not seem that the upstream issue will be fixed any time soon due to binary compatibility concerns. Nevertheless, it makes no sense, and it is not possible possible to create a *value* of a "package type".

Therefore, we work around the issue by tolerating a `PackageRef` as the type of a PARAM of an inline accessor method. We replace it by `Nothing`, which matches the fact that no value can be created for it.

Note that the *signature* and therefore the *signed name* of such a method will be different from what dotc computes, so it will still be semantically broken.


adpi2 (approved):
Nice

PR #433, opened 2023-12-22
Make the entire API thread-safe.

sjrd:
On the condition that the `ClasspathEntry`s used to create the `Context` are thread-safe.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
@@ -2,6 +2,9 @@ package tastyquery
 
 import scala.annotation.{switch, tailrec}
 
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.{AtomicBoolean, AtomicReference}
+
 import scala.collection.mutable
 
 import tastyquery.Annotations.*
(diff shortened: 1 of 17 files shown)


bishabosha (approved):

tasty-query/shared/src/main/scala/tastyquery/Symbols.scala
```diff
@@ -1706,24 +1730,43 @@ object Symbols {
...
     import PackageSymbol.*
+    import tastyquery.reader.Loaders.PackageLoadingInfo
 
     type DefiningTreeType = Nothing
     type DeclType = Symbol
 
     private[Symbols] val specialKind: SpecialKind = computeSpecialKind(name, owner)
 
+    /** Package loading info with raw data from the classpath. */
+    private var optLoadingInfo: Option[PackageLoadingInfo] = None
```
bishabosha, line 1741: just checking why this is safe to be a plain var?

bishabosha: im assuming that the flow is that it is the thread that initialised the context is the thread that both creates this object and sets this, but then when is the update of setting this field  "published" to other threads

sjrd: Hum, good question. For the members of other `Symbol`s, they should be published thanks to `myDeclarations.compareAndSet` at their `PackageSymbol`'s level. But for the things initialized during context initialization, we don't have a definitive release barrier. We might need to add one at the end of `Context.initialize` perhaps?

bishabosha: could be good

sjrd: Done.

PR #435, opened 2024-01-03
Fix source file name on Windows

### LINKED ISSUE #434: `SourceFile.name` returns full path on Windows

adpi2:
Fix #434


adpi2:
Actually that's wrong: on Windows we should still be able to read Unix-style path.

PR #454, opened 2025-05-09
Upgrade to Scala 3.7.1.

mrdziuban:
Upgrades to Scala 3.7.0 and 2.13.16 for Scala 2 tests. Also upgrades Scala.js to 1.19.0 to match the compiler upgrade in https://github.com/scala/scala3/pull/23026.

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/reader/classfiles/ClassfileParser.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/reader/classfiles/ClassfileParser.scala
@@ -132,7 +132,7 @@ private[reader] object ClassfileParser {
         .use(ClassfileReader.readAnnotation(Set(annot.ScalaLongSignature, annot.ScalaSignature)))
         .getOrElse(failNoAnnot())
 
-    val sigBytes = scalaSigAnnotation.tpe match {
+    val sigBytes = (scalaSigAnnotation.tpe: @unchecked) match {
       case annot.ScalaSignature =>
         val bytesArg = scalaSigAnnotation.values.head._2.asInstanceOf[AnnotationValue.Const]
         pool.sigbytes(bytesArg.valueIdx)
--- a/tasty-query/shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
@@ -31,9 +31,9 @@ private[pickles] class PickleReader {
   private var frozenSymbols: Boolean = false
 
   /** The map from created local symbols to the address of their info, until it gets read. */
-  private val localSymbolInfoRefs = mutable.AnyRefMap.empty[TermOrTypeSymbol, Int]
+  private val localSymbolInfoRefs = mutable.HashMap.empty[TermOrTypeSymbol, Int]
 
-  private val localClassGivenSelfTypeRefs = mutable.AnyRefMap.empty[ClassSymbol, Int]
+  private val localClassGivenSelfTypeRefs = mutable.HashMap.empty[ClassSymbol, Int]
 
   final class Structure(using val myEntries: Entries, val myIndex: Index):
     def allRegisteredSymbols: Iterator[TermOrTypeSymbol] =
--- a/build.sbt
+++ b/build.sbt
@@ -3,7 +3,7 @@ import sbt.internal.util.ManagedLogger
 
 import org.scalajs.jsenv.nodejs.NodeJSEnv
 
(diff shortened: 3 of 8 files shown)


tasty-query/shared/src/main/scala/tastyquery/reader/classfiles/ClassfileParser.scala
```diff
@@ -140,6 +140,7 @@ private[reader] object ClassfileParser {
         val bytesArrArg = scalaSigAnnotation.values.head._2.asInstanceOf[AnnotationValue.Arr]
         val idxs = bytesArrArg.values.map(_.asInstanceOf[AnnotationValue.Const].valueIdx)
         pool.sigbytes(idxs)
+      case _: SimpleName => failNoAnnot()
```
mrdziuban, line 143: This was reported as an inexhaustive match. I figured failing seemed reasonable but I'm happy to change it if you think there's a better option @sjrd

tasty-query/shared/src/main/scala/tastyquery/reader/pickles/PickleReader.scala
```diff
@@ -31,9 +31,9 @@ private[pickles] class PickleReader {
   private var frozenSymbols: Boolean = false
 
   /** The map from created local symbols to the address of their info, until it gets read. */
-  private val localSymbolInfoRefs = mutable.AnyRefMap.empty[TermOrTypeSymbol, Int]
+  private val localSymbolInfoRefs = mutable.HashMap.empty[TermOrTypeSymbol, Int]
```
mrdziuban, line 34: `mutable.AnyRefMap` was deprecated in https://github.com/scala/scala/pull/10862 and now encourages use of `mutable.HashMap`

mrdziuban:
Happy to address the version policy failures, just not sure what the best way to do so is. Let me know

sjrd:
> Happy to address the version policy failures, just not sure what the best way to do so is. Let me know

See for example #449.

mrdziuban:
> See for example #449.

Thanks @sjrd. I was able to get it working by setting `versionPolicyIntention := Compatibility.None`. `Compatibility.BinaryCompatible` didn't work like in #449 due to the change from `AnyRefMap` to `HashMap`

sjrd:
> > See for example #449.
> 
> Thanks @sjrd. I was able to get it working by setting `versionPolicyIntention := Compatibility.None`. `Compatibility.BinaryCompatible` didn't work like in #449 due to the change from `AnyRefMap` to `HashMap`

These are all private fields. They should be MiMa-filtered instead.

mrdziuban:
> > > See for example #449.
> > 
> > 
> > Thanks @sjrd. I was able to get it working by setting `versionPolicyIntention := Compatibility.None`. `Compatibility.BinaryCompatible` didn't work like in #449 due to the change from `AnyRefMap` to `HashMap`
> 
> These are all private fields. They should be MiMa-filtered instead.

Okay, I can do that, but what do you suggest for this error when running `tastyQueryJS/versionPolicyCheck` due to the Scala.js upgrade?

```
Incompatibilities with dependencies of tasty-query:1.5.0
  org.scala-js:scalajs-scalalib_2.13: incompatible version change from 2.13.13+1.16.0 to 2.13.16+1.19.0 (compatibility: strict semantic versioning)
```

mrdziuban:
Just following up here @sjrd, do you have any suggestions for the Scala.js incompatibility issue?

mrdziuban:
Hi @sjrd, wanted to follow up again. Now that the [Scala 3.7.1 release](https://github.com/scala/scala3/issues/23301) is in progress I would love to push this over the finish line. For context, I'm trying to unblock the upgrade in [a tool I built on top of tasty-query](https://github.com/mblink/find-unused).

What can I do to work around the reported Scala.js incompatibility issue?

bishabosha:
@mrdziuban it should be a new minor release anyway so version compat should be none

bishabosha:
if the message is something so opaque as 
```
[error] Incompatibilities with dependencies of tasty-query:1.5.0
[error]   org.scala-js:scalajs-scalalib_2.13: incompatible version change from 2.13.13+1.16.0 to 2.13.16+1.19.0 (compatibility: strict semantic versioning)
```
should that imply that the only solution is the `versionPolicyIntention := None`?

sjrd (approved):

sjrd:
Version 1.6.0 is on its way to Maven Central.

Thanks for pushing this, and sorry for the lack of responsivity in the past few weeks.

mrdziuban:
Thank you!

PR #463, opened 2026-01-27
Re-enable `-source:future`

mrdziuban:
Re-enables `-source:future` (which was disabled in #461) and fixes all warnings. There were four types of changes necessary:

1. Change `implicit` to `using`
2. Change `if (pred) ... else ...` to `if pred then ... else ...`
3. Change `while (pred) ...` to `while pred do ...`
4. Change `implicit def` conversions to instances of `Conversion`


mrdziuban:
Closing due to lack of activity, feel free to update and merge if you feel it's worthwhile

PR #465, opened 2026-03-23
Handle named args in pattern matching

Krever:
Hey 👋 I spotted the bug in one of my projects and asked Claude to fix it. 

Disclaimer: I have zero clue about the bigger picture here, let me know if there is more that should be considered in fixing this. (The test I added fails without the fix)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -37,7 +37,7 @@ inThisBuild(Def.settings(
     Developer("sjrd", "Sébastien Doeraene", "sjrdoeraene@gmail.com", url("https://github.com/sjrd/")),
     Developer("bishabosha", "Jamie Thompson", "bishbashboshjt@gmail.com", url("https://github.com/bishabosha")),
   ),
-  versionPolicyIntention := Compatibility.BinaryAndSourceCompatible,
+  versionPolicyIntention := Compatibility.BinaryCompatible,
   // Ignore dependencies to internal modules whose version is like `1.2.3+4...` (see https://github.com/scalacenter/sbt-version-policy#how-to-integrate-with-sbt-dynver)
   versionPolicyIgnoredInternalDependencyVersions := Some("^\\d+\\.\\d+\\.\\d+\\+\\d+".r)
 ))
--- a/tasty-query/shared/src/main/scala/tastyquery/reader/tasties/TreeUnpickler.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/reader/tasties/TreeUnpickler.scala
@@ -897,6 +897,11 @@ private[tasties] class TreeUnpickler private (
       reader.readByte()
       val shared = forkAt(reader.readAddr()).readPattern
       if spn.isUnknown then shared else shared.withPos(spn)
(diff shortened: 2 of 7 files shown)


Krever:
Hey @sjrd, could you take a look maybe?

tasty-query/shared/src/main/scala/tastyquery/reader/tasties/TreeUnpickler.scala
```diff
@@ -897,6 +897,11 @@ private[tasties] class TreeUnpickler private (
       reader.readByte()
       val shared = forkAt(reader.readAddr()).readPattern
       if spn.isUnknown then shared else shared.withPos(spn)
+    case NAMEDARG =>
+      // Named extractor patterns (e.g., `case Foo(field = value)`) wrap sub-patterns in NAMEDARG; discard the name
+      reader.readByte()
+      readUnsignedName()
+      readPattern
```
sjrd: This entirely discards the name. That does not seem right. If the info is in TASTy, we should surface it in the tasty-query tree.

sjrd (changes requested):

tasty-query/shared/src/main/scala/tastyquery/reader/tasties/TreeUnpickler.scala
Krever: Good catch, I added the tree node to carry the name. Is that what you meant?

build.sbt
```diff
@@ -126,6 +126,10 @@ lazy val tastyQuery =
       mimaBinaryIssueFilters ++= {
         import com.typesafe.tools.mima.core.*
         Seq(
+          // New pattern tree node for named extractor patterns `case Foo(field = pat)`
+          ProblemFilters.exclude[MissingClassProblem]("tastyquery.Trees$NamedPattern"),
+          ProblemFilters.exclude[MissingClassProblem]("tastyquery.Trees$NamedPattern$"),
+          ProblemFilters.exclude[MissingFieldProblem]("tastyquery.Trees.NamedPattern"),
```
sjrd: The appropriate way to deal with that is to set the version policy for `BinaryCompatible` to prepare a minor release. We do this with a separate commit like the following:
https://github.com/scalacenter/tasty-query/commit/e4b3a5092edc2e3aec67e0f8d1d9fcc39075032f
so the PR would have 2 commits:

1. a commit to change the version policy
2. a commit with the actual changes.

Do you prefer to adapt the PR yourself, or that I do it?

sjrd (commented):
The changes LGTM. There's a process issue with the MiMa exclusions.

build.sbt
Krever: I think it might be simpler if you do it – it would be the first time for me to interact with MiMA and the process, so I might be introducing more issues. Having said that, I’m happy to give it a try if it's too much load on your side.

sjrd: No that's OK. I'll do that later today.

sjrd (approved):

PR #469, opened 2026-05-07
Fix incorrect spans assigned in some cases

jadenPete:
I've observed tasty-query occasionally reading the wrong position for certain trees.

Consider this example:
```scala
class InferredTypeArgs {
  def id[T](x: T): T = x

  val a = id(42)
}
```

Here, the compiler inserts an inferred type annotation, so in TASTy, the `ValDef` statement looks like this:
```scala
val a: scala.Int = id[scala.Int](42)
```

The `scala.Int` type annotation should have a zero-width span because it doesn't exist in the source text, but tasty-query incorrectly reads its span as the same as that of `id(42)`. Also, the `scala.Int` type argument should have the same span as the function it's applied to (`id`), but it's instead the same as that of the entire function call (`id(42)`).

This PR fixes both of these issues. I've verified that the tests I added fail before the fix, and succeed after the fix.

### DIFF
--- a/tasty-query/shared/src/test/scala/tastyquery/PositionSuite.scala
+++ b/tasty-query/shared/src/test/scala/tastyquery/PositionSuite.scala
@@ -309,6 +309,37 @@ class PositionSuite extends RestrictedUnpicklingSuite {
     assertEquals(collectCode[TypeApply](tree), List("Seq", "Seq[Int]", "A[Int, Seq[String]]"))
   }
 
+  testUnpickle("inferred-type-arg", "simple_trees.InferredTypeArgs") { tree =>
+    /*
+     * The inferred `Int` type argument in `id(42)` is encoded as a non-tree-tagged type (a `TYPEREF`). When tasty-query
+     * reads it, it must capture `span` before consuming the type bytes, otherwise the resulting `TypeWrapper` gets the
+     * position of whatever node comes next.
+     */
+    val idName = termName("id")
+    val typeApply = findTree(tree) {
+      case typeApply @ TypeApply(Select(_, SignedName(`idName`, _, _)), _) => typeApply
+      case typeApply @ TypeApply(Select(_, `idName`), _)                   => typeApply
+    }
+
+    val typeArg = typeApply.args match {
+      case (typeWrapper: TypeWrapper) :: Nil => typeWrapper
+      case other                             => fail(s"expected a single TypeWrapper, got: $other")
+    }
+
+    assertEquals(posToCode(typeArg.pos), Some("id"))
+  }
+
+  testUnpickle("package-ref-ident", "simple_trees.PackageRefIdent") { tree =>
+    /*
+     * The leading `scala` in `scala.Predef.identity(1)` is an `Ident` with a `TermRef` type pickled as TERMREFpkg. The
+     * unpickler must use the span captured before `readPackageRef` advances the reader, otherwise the `Ident` inherits
+     * the position of the next entry.
+     */
+    val scalaIdentifier = findTree(tree) { case identifier @ Ident(SimpleName("scala")) => identifier }
+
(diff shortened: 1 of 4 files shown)


tasty-query/shared/src/test/scala/tastyquery/PositionSuite.scala
```diff
@@ -369,12 +396,6 @@ class PositionSuite extends RestrictedUnpicklingSuite {
       collectCode[TypeDefinitionTree](tree),
       List(
         "Int",
-        /* The following makes no sense; it is the position of the type def
-         * of `type AbstractType`. It's probably dotc's auto-assigning of
-         * positions that goes wild.
-         */
-        """type AbstractType
-          |  type AbstractWithBounds >: Null <: Product""".stripMargin,
```
sjrd, line 377: I find it suspicious that this item *disappears*, instead of being fixed. Same thing below. Could you explain why that happens?

tasty-query/shared/src/test/scala/tastyquery/PositionSuite.scala
```diff
@@ -309,6 +309,33 @@ class PositionSuite extends RestrictedUnpicklingSuite {
     assertEquals(collectCode[TypeApply](tree), List("Seq", "Seq[Int]", "A[Int, Seq[String]]"))
   }
 
+  testUnpickle("inferred-type-arg", "simple_trees.InferredTypeArgs") { tree =>
+    // The inferred `Int` type argument in `id(42)` is encoded as a non-tree-tagged type (a `TYPEREF`). When tasty-query
+    // reads it, it must capture `span` before consuming the type bytes, otherwise the resulting `TypeWrapper` gets the
+    // position of whatever node comes next.
```
sjrd: Nit: we prefer the following style for multiline comments:

```suggestion
    /* The inferred `Int` type argument in `id(42)` is encoded as a non-tree-tagged type (a `TYPEREF`). When tasty-query
     * reads it, it must capture `span` before consuming the type bytes, otherwise the resulting `TypeWrapper` gets the
     * position of whatever node comes next.
     */
```
Same below.

sjrd (commented):
Very nice. Thank you. I have some comments.

tasty-query/shared/src/test/scala/tastyquery/PositionSuite.scala
jadenPete: I think it's because of the `!pos.isZeroExtent` check here:
https://github.com/scalacenter/tasty-query/blob/main/tasty-query/shared/src/test/scala/tastyquery/PositionSuite.scala#L27

The type has no bounds, so the returned string is empty. If I remove that check, I get this error:
```
==> X tastyquery.PositionSuite.type-definition-tree-1  0.011s munit.ComparisonFailException: /home/jpeterson/lucid/tasty-query/tasty-query/shared/src/test/scala/tastyquery/PositionSuite.scala:404
403:      )
404:    )
405:  }
values are not the same
=> Obtained
List(
  "Int",
  "",
  ">: Null <: Product",
  "Int",
  "Null <: Product = Null",
  "Null <: Product"
)
=> Diff (- obtained, + expected)
   "Int",
-  "",
   ">: Null <: Product",
    at munit.Assertions.failComparison(Assertions.scala:278)
```

tasty-query/shared/src/test/scala/tastyquery/PositionSuite.scala
jadenPete: Fixed!

tasty-query/shared/src/test/scala/tastyquery/PositionSuite.scala
sjrd: Ah that makes sense.

sjrd (approved):

PR #475, opened 2026-09-03
Do not eagerly create the scala.caps package

### LINKED ISSUE #474: AssertionError "trying to add a second entry symbol[scala.caps]" loading a Scala 3.2–3.6 stdlib
## Summary

Loading a scala3-library_3 jar from Scala 3.2 through 3.6 and enumerating the scala package throws an AssertionError. In those versions scala.caps is an object, but Definitions eagerly creates a package symbol of that name, so unpickling scala/caps.tasty tries to add a second caps entry to package scala.

## Reproduction

```scala
  //> using scala 3.9.0
  //> using dep ch.epfl.scala::tasty-query:1.9.0
  import tastyquery.Contexts.*, tastyquery.jdk.ClasspathLoaders

  @main def run(entries: String*) =
    val ctx: Context = Context.initialize(
      ClasspathLoaders.read(entries.toList.map(java.nio.file.Path.of(_)))
    )
    given Context = ctx
    println(ctx.findPackage("scala").declarations.map(_.name.toString).sorted)
```
```
Exception in thread "main" java.lang.AssertionError: assertion failed: trying to add a second entry symbol[scala.caps] for name caps in symbol[package <root>.scala]
      at tastyquery.Symbols$PackageSymbol.addDecl(Symbols.scala:1833)
      at tastyquery.Symbols$Symbol.addDeclIfDeclaringSym(Symbols.scala:145)
      at tastyquery.Symbols$TermSymbol$.create(Symbols.scala:670)
      at tastyquery.reader.tasties.TreeUnpickler.createSymbols(TreeUnpickler.scala:108)
      ...
      at tastyquery.reader.Loaders$PackageLoadingInfo.loadAllRoots(Loaders.scala:46)
      at tastyquery.Symbols$PackageSymbol.declarations(Symbols.scala:1932)
```

Note findTopLevelClass("scala.CanEqual") on the same context still succeeds — only enumerating the package's declarations fails.

## Affected versions

Regression introduced in 1.7.0, against the repro above:

1.6.1 ok
1.7.0 AssertionError
1.8.0 AssertionError
1.9.0 AssertionError

## Affected scala3-library_3 versions — exactly the window where scala.caps exists and is an object:

3.0.x, 3.1.x - absent scala.caps - ok         
3.2.x – 3.6.x - an object (scala/caps.class, scala/caps$.class, scala/caps.tasty) - AssertionError
3.7.x+ - a package (scala/caps/Capability.tasty, …) - ok

## Cause

Definitions.scala:35 creates the package unconditionally at context initialization:

private val capsPackage = scalaPackage.getPackageDeclOrCreate(termName("caps"))

Unpickling a pre-3.7 scala/caps.tasty then adds a term symbol caps to package scala, which already holds that package, and PackageSymbol.addDecl asserts on the duplicate name.

## Prior art

Given the same collision, dotc reports it as a warning and continues:

package scala contains object and package with same name: caps.
This indicates that there are several versions of the Scala standard library on the classpath.

rochala:
Fixes #474

The one thing missing here is test coverage. I decided that it is not worth to implement it as it will require us to either create a test source that is built using Scala <3.0, ..., 3.6> because afterwards it resulted in compilation error thus we can't recreate it without new module or some other hacks. Other way we could test it is to bundle one of <3.2, ..., 3.6> stdlibs and just try to read from it.

As mentioned in AI policy:
Implementation done by Fable 5.1, Comments revised and created by me left for the future contributors unless you don't want it.

Also it breaks MiMa because PureClass is now Option, release will require version bump I think ?

### DIFF
--- a/tasty-query/shared/src/main/scala/tastyquery/Definitions.scala
+++ b/tasty-query/shared/src/main/scala/tastyquery/Definitions.scala
@@ -32,7 +32,6 @@ final class Definitions private[tastyquery] (
   private val javaPackage = RootPackage.getPackageDeclOrCreate(nme.javaPackageName)
   val javaLangPackage = javaPackage.getPackageDeclOrCreate(nme.langPackageName)
   private[tastyquery] val javaLangInvokePackage = javaLangPackage.getPackageDeclOrCreate(termName("invoke"))
-  private val capsPackage = scalaPackage.getPackageDeclOrCreate(termName("caps"))
 
   private val scalaAnnotationPackage =
     scalaPackage.getPackageDeclOrCreate(termName("annotation"))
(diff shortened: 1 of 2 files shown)


rochala:
@sjrd Do I just bump it to 1.10 or do you want me to add an unsafe member to retain binary compat and use new MaybePureClass ?

tasty-query/shared/src/main/scala/tastyquery/Definitions.scala
```diff
@@ -498,7 +497,21 @@ final class Definitions private[tastyquery] (
 
   lazy val ProductClass = scalaPackage.requiredClass("Product")
 
-  lazy val PureClass = capsPackage.requiredClass("Pure")
+  /** `scala.caps.Pure`, depending on the standard library on the classpath:
+    *   - Scala 3.0 and 3.1: `scala.caps` does not exist, so `None`
+    *   - Scala 3.2 to 3.6: `scala.caps` is an object and `Pure` a member of it
+    *   - Scala 3.7 and later: `scala.caps` is a package
+    */
+  lazy val PureClass: Option[ClassSymbol] = withRestrictedContext {
```
sjrd: That's a major breaking change.

Keep returning `ClassSymbol` and throw if it's not there.

tasty-query/shared/src/main/scala/tastyquery/Definitions.scala
```diff
@@ -498,7 +497,24 @@ final class Definitions private[tastyquery] (
...
+    def fromObject = scalaPackage
+      .getDecl(termName("caps"))
+      .collect { case caps: TermSymbol => caps }
+      .flatMap(_.moduleClass)
+      .flatMap(_.getDecl(typeName("Pure")))
+    // Package first: `getDecl` loads the `caps` root, which asserts if a `caps` package already exists.
+    fromPackage.orElse(fromObject).map(_.asClass)
+  }
+
+  lazy val PureClass: ClassSymbol =
```
rochala, line 516: We have to add new member because all intersection types on 3.0 and 3.1 would throw

sjrd (approved):

PR #476, opened 2026-09-08
Implement Symbol.docComment returning docstring and its span

rochala:
Original implementation in dotc:
```scala
package dotty.tools.dotc
package core.tasty

import core.Comments.Comment
import util.Spans.Span
import util.HashMap

import dotty.tools.tasty.{TastyReader, TastyBuffer}
import TastyBuffer.Addr

import java.nio.charset.StandardCharsets

class CommentUnpickler(reader: TastyReader) {
  import reader.*

  private[tasty] lazy val comments: HashMap[Addr, Comment] = {
    val comments = new HashMap[Addr, Comment]
    while (!isAtEnd) {
      val addr = readAddr()
      val rawComment = readUtf8()
      val position = new Span(readLongInt())
      comments(addr) = Comment(position, rawComment)
    }
    comments
  }

  def commentAt(addr: Addr): Option[Comment] =
    comments.get(addr)
}
```

We put it in immutable collection + we use SourcePosition vs Span, if you want it 1-1 aligned let me know, ~~but I think this is better shape~~ (Actually lets keep it aligned, SourceFile can be easily extracted via `Symbol.tree.get.pos.sourceFile`).
Written with help of Claude

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -37,7 +37,7 @@ inThisBuild(Def.settings(
     Developer("sjrd", "Sébastien Doeraene", "sjrdoeraene@gmail.com", url("https://github.com/sjrd/")),
     Developer("bishabosha", "Jamie Thompson", "bishbashboshjt@gmail.com", url("https://github.com/bishabosha")),
   ),
-  versionPolicyIntention := Compatibility.BinaryAndSourceCompatible,
+  versionPolicyIntention := Compatibility.BinaryCompatible,
   // Ignore dependencies to internal modules whose version is like `1.2.3+4...` (see https://github.com/scalacenter/sbt-version-policy#how-to-integrate-with-sbt-dynver)
   versionPolicyIgnoredInternalDependencyVersions := Some("^\\d+\\.\\d+\\.\\d+\\+\\d+".r)
 ))
@@ -127,6 +127,8 @@ lazy val tastyQuery =
       mimaBinaryIssueFilters ++= {
         import com.typesafe.tools.mima.core.*
         Seq(
+          // constructor of a class nested in a private[reader] object; no issue
+          ProblemFilters.exclude[DirectMissingMethodProblem]("tastyquery.reader.tasties.TastyUnpickler#TreeSectionUnpickler.this"),
         )
       },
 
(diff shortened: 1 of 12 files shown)


tasty-query/shared/src/main/scala/tastyquery/Spans.scala
```diff
@@ -167,6 +167,9 @@ private[tastyquery] object Spans {
   /** A synthetic zero-extent span that starts and ends at given `start`. */
   def Span(start: Int): Span = Span(start, start)
 
+  object Span:
+    def fromCoords(coords: Long): Span = new Span(coords)
```
sjrd: I'm not of a fan of the `object Span` that overloads `def Span`s. Admittedly the latter should have been `def apply`'s in `object Span`, but oh well.

Prefer a `def spanFromCoord` instead.

build.sbt
```diff
@@ -37,7 +37,7 @@ inThisBuild(Def.settings(
     Developer("sjrd", "Sébastien Doeraene", "sjrdoeraene@gmail.com", url("https://github.com/sjrd/")),
     Developer("bishabosha", "Jamie Thompson", "bishbashboshjt@gmail.com", url("https://github.com/bishabosha")),
   ),
-  versionPolicyIntention := Compatibility.BinaryAndSourceCompatible,
+  versionPolicyIntention := Compatibility.BinaryCompatible,
```
sjrd, line 40: Please make a separate, first commit for this change, like this commit:
https://github.com/scalacenter/tasty-query/commit/d3628abd21623255ad6934de455399ef35f1b95d

tasty-query/shared/src/test/scala/tastyquery/DocstringSuite.scala
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  testWithContext("local-definition") {
+    val Docstrings = ctx.findTopLevelClass("simple_trees.Docstrings")
+    val withLocal = Docstrings.findNonOverloadedDecl(termName("withLocal"))
+    object collector extends TreeTraverser:
+      var localSymbol: Option[TermSymbol] = None
+      override def traverse(tree: Tree): Unit =
+        tree match
+          case ValDef(SimpleName("local"), _, _, symbol) => localSymbol = Some(symbol)
```
sjrd: We have `TestUtils.findValDef` for this.

sjrd (changes requested):
The MiMa failure seems to be about something transitively private. It can be filtered out.

rochala:
Applied comments + as I mentioned in PR description, I removed source file from DocComment to instead rely on span as you already have other way to access it

rochala:
Any chance to get a re-review when you find a while :D ?

PR #477, opened 2026-09-08
Update sbt-release-ci to fix worktree support + bump sbt

rochala:
While working on those issuses I found that worktree support is broken. It is because of sbt-release-ci plugin that tried to use JGit < 7.0 which does not support worktrees.

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.11.3
+sbt.version=1.13.0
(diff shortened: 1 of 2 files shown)


sjrd:
What's "worktree support"?

rochala:
`git worktree add ../new-worktree-name`
https://git-scm.com/docs/git-worktree

I'm leaving docs here ^^
But tldr a git repo can have multiple working trees, allowing you to check out more than one branch at a time. Then you can have few different branches that you can work on at the same time without cloning whole repository.

For example I have 2 worktrees for tasty query at the moment all based on single git repo:
image

sjrd (approved):

